PR #14, opened 2021-09-01
Element

james-d-mitchell:


james-d-mitchell:
I need to make another release of libsemigroups before this will pass.

PR #19, opened 2021-10-06
add pkgconfig and packaging as build dependencies

chinmaya1011:


james-d-mitchell:
@chinmaya1011 good, I think if you rebase onto the most recent `main` branch, then the CI will pass.

PR #27, opened 2021-11-11
Add installation from conda instructions

james-d-mitchell:


james-d-mitchell:
This will fail until the docker container for libsemigroups v2.0.3 is available.

PR #36, opened 2021-11-23
fix pkg_config versions and add long_description 

### LINKED ISSUE #33: Fix pkgconfig version numbers and add to installation requirements
Required versions for the `pkgconfig` python package are based are based on version numbers of the pkg-config program. It should also be added as an install requirement because it is used in tools.py.

### LINKED ISSUE #35: add long_description
Contents show up as README for PyPI package.

https://packaging.python.org/guides/making-a-pypi-friendly-readme/

chinmayanagpal:
closes #33 by adding pkgconfig>=1.5.0 as a setup and installation requirement
closes #35 by setting long_description to the contents of the README.md file


pyproject.toml
```diff
@@ -5,7 +5,7 @@ requires = [
     "setuptools>=42",
     "wheel",
     "pybind11>=2.7.0",
-    "pkgconfig>=0.29.2",
+    "pkgconfig>=l.5.0",
```
james-d-mitchell, line 8: That looks suspiciously like an "l" not a "1", is this really correct?

james-d-mitchell: Happy to approve and merge mod this one typo!

chinmayanagpal: ah sorry!

PR #39, opened 2021-11-26
add Build workflow

chinmayanagpal:


chinmayanagpal:
@james-d-mitchell  is this ok?

james-d-mitchell:
@chinmayanagpal this looks pretty amazing. Two comments

1) we definitely don't want to be producing wheels:
~~~
on: [push, pull_request, workflow_dispatch]
~~~
but only on `release` I think. I'm not sure what `workflow_dispatch` is, so maybe we want to keep this too.

2) It looks like you need to rebase this on `main`, because at the moment your PR contains some commits that are already in `main`, and shouldn't appear again. You should be able to do this in your `build` branch by doing `git rebase main`.

chinmayanagpal:
Sorry for the force pushes, there doesn't seem to be any way to test GitHub Actions locally. Right now I can't figure out how to put the version of libsemigroups it installs as a variable so it's just written into the download commands on two separate lines. Should I go with this approach? The other option is to just get the latest release (I think).

chinmayanagpal:
@james-d-mitchell here's the latest run of the build workflow, could you take a look? https://github.com/chinmayanagpal/libsemigroups_pybind11/runs/4396648848 

I think it'll have to run on the libsemigroups runner for the upload job to work, but other than that, I just need to change the options to only run it on releases and remove the CIBW_BUILD line that tells it to only build for specific python versions (I added it so I'd be able to test the workflow without waiting two hours).

PR #56, opened 2022-09-10
Rc v0.2.1

james-d-mitchell:


james-d-mitchell:
This seems to require a new release of `libsemigroups` to remove the return type `auto` from `longest_common_subword` (I think this is the issue here).

Issue #65, opened 2022-10-02
Remove non-pytest tests from tests


james-d-mitchell:
Basically convert the test files to not use `self.assertEqual` etc

james-d-mitchell:
Resolved in v0.4.0

Issue #63, opened 2022-10-01
Add support for `NEGATIVE_INFINITY` in `libsemigroups`


james-d-mitchell:

james-d-mitchell:
Resolved in v0.4.0

PR #66, opened 2022-10-02
Add support for constants and for powering some types of matrix

james-d-mitchell:


james-d-mitchell:
Resolves #63 and #64

james-d-mitchell:
And resolves #65

Issue #67, opened 2022-10-03
Add some more CI jobs


james-d-mitchell:
Testing different versions of python and different os's.

james-d-mitchell:
Resolved by #105.

PR #105, opened 2023-01-12
ci: test against python 3.8-3.11

james-d-mitchell:


james-d-mitchell:
Will resolve issue #67 if it works

PR #108, opened 2023-01-24
Update tests for monoid output for orientation preserving and reversing monoids

MTWhyte:
This PR updates the tests for `orientation_preserving_monoid` and `orientation_reversing_monoid` to work with monoid presentation output, rather than semigroup presentation.

### DIFF
--- a/tests/test_fpsemi_examples.py
+++ b/tests/test_fpsemi_examples.py
@@ -493,11 +493,15 @@ def test_singular_brauer_monoid_exceptions():
 
 
 def test_orientation_preserving_monoid():
-    s = orientation_preserving_monoid(6)
+    p = make(orientation_preserving_monoid(6))
+    p.alphabet(3)
+    presentation.replace_word(p, [], [2])
+    presentation.add_identity_rules(p, 2)
+    p.validate()
     tc = ToddCoxeter(congruence_kind.twosided)
     tc.set_number_of_generators(3)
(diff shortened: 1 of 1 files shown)


MTWhyte:
The CI tanked, but we expected this, right?

james-d-mitchell:
Thanks @MTWhyte, yes I think I'd expect the CI to tank. Perhaps the test can be made more robust using something similar to:

https://github.com/libsemigroups/libsemigroups_pybind11/blob/ef498fb7fa77555846d86c2dfb61734b9adc48c5/tests/test_present.py#L631-L641


This way we can merge and everything will be ok!

MTWhyte:
> Thanks @MTWhyte, yes I think I'd expect the CI to tank. Perhaps the test can be made more robust using something similar to:
> 
> https://github.com/libsemigroups/libsemigroups_pybind11/blob/ef498fb7fa77555846d86c2dfb61734b9adc48c5/tests/test_present.py#L631-L641
> 
> This way we can merge and everything will be ok!

I think we touched on this the other day, I understand `presentation.make` gives a string presentation, whereas `fpsemigroup.make` gives a word-type presentation. Just to check: since I'm using the latter here, should this test be independent of the version? Or am I mistaken and in fact `make` from both namespaces will use strings?

james-d-mitchell:
Now superseded by other developments

PR #118, opened 2023-02-06
Add functionality for `ToddCoxeter` function `to_gap_string`

MTWhyte:
This PR adds functionality for the `ToddCoxeter` function `to_gap_string`.

I think the CI will fail until there is a release of `libsemigroups` which includes the PR which adds argument checking for the sidedness of the congruence.


james-d-mitchell:
Thanks @MTWhyte, please add some if statements to the tests to test the version of libsemigroups and test the behaviour if the version is 2.6.0 or higher, as mentioned here:

https://github.com/libsemigroups/libsemigroups_pybind11/pull/108

Issue #100, opened 2023-01-06
More colours for `ActionDigraph` `graphviz` representations


MTWhyte:
It would be nice to have more edge colours available for the `graphviz` representations of `ActionDigraph`. There are five colours currently available.

Rather than just extending the colour palette, I think the most robust thing to do might be to have a means of generating an arbitrary number of 'distinct' colours, perhaps via the information [here](https://stackoverflow.com/questions/470690/how-to-automatically-generate-n-distinct-colors).

On the less important aesthetic front, there's a lightweight python package called [`palettable`](https://jiffyclub.github.io/palettable/) which contains a whole load of nice-looking colour schemes (we'd want the 'qualitative' ones). This is perhaps also a viable way to add in colours -- some of these palettes have quite a lot of colours in them, and arguably we'd not be that interested in getting pictures of `ActionDigraph` instances with arbitrary large degrees.

james-d-mitchell:
This is a great idea, I'd welcome a PR doing this!

MTWhyte:
Better yet for contrast and accessibility are the colour schemes at https://personal.sron.nl/~pault/ (thanks to @flsmith for indirectly pointing this out)

MTWhyte:
For reference when I make tests for this: the behaviour I find on the current `main` is:

- Where there are less than or equal to four distinct edge colours, things are coloured as expected
- Where there are five distinct edge colours, everything is coloured in black and white
- Where there are six or more distinct edge colours, an exception is thrown

It would be pretty cool to be able to catch the black and white case in a test, but I can't immediately think of how.

james-d-mitchell:
Resolved in v0.9.0

PR #125, opened 2023-03-05
Improve `ActionDigraph` `graphviz` colour capability

MTWhyte:
This PR improves the `graphviz` colour capability for `ActionDigraph` objects, as discussed in issue #100.

More specifically: a nice palate of nine colourblind accessible colours due to [Paul Tol](https://personal.sron.nl/~pault/) is used. Where a graph has more than nine distinct labels whose edges need coloured, these nine colours are simply reused cyclically. This seemed like a reasonable way of doing things.

### DIFF
--- a/tests/test_actiondigraph.py
+++ b/tests/test_actiondigraph.py
@@ -872,7 +872,7 @@ def test_dot():
     d = action_digraph_helper.make(3, l)
     assert (
         str(action_digraph_helper.dot(d))
-        == 'digraph {\n\tnode [shape=circle]\n\t0\n\t1\n\t2\n\t0 -> 0 [label=0 color="#ee1487"]\n\t0 -> 1 [label=1 color="#0dda4"]\n\t1 -> 1 [label=0 color="#ee1487"]\n\t1 -> 0 [label=1 color="#0dda4"]\n\t2 -> 2 [label=0 color="#ee1487"]\n\t2 -> 2 [label=1 color="#0dda4"]\n}\n'  # pylint: disable=line-too-long
+        == 'digraph {\n\tnode [shape=circle]\n\t0\n\t1\n\t2\n\t0 -> 0 [label=0 color="#cc6677"]\n\t0 -> 1 [label=1 color="#ddcc77"]\n\t1 -> 1 [label=0 color="#cc6677"]\n\t1 -> 0 [label=1 color="#ddcc77"]\n\t2 -> 2 [label=0 color="#cc6677"]\n\t2 -> 2 [label=1 color="#ddcc77"]\n}\n'  # pylint: disable=line-too-long
     )
 
     l = list(12 * [i] for i in range(4))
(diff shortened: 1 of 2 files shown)


MTWhyte:
Note that test I have is currently failing, or at least it was on my computer before I pushed. This is due to new lines in the string corresponding to the `dot` object. I've tried to do this in a similar way to the previous test, but it is not working (possibly this is because the string is quite long?), and I don't know how to get it to behave.

tests/test_actiondigraph.py
```diff
@@ -874,3 +874,11 @@ def test_dot():
...
...e"]\n\t1 -> 1 [label=4 color="#44aa99"]\n\t1 -> 1 [label=5 color="#882255"]\n\t1 -> 1 [label=6 color="#44aa99"]\n\t1 -> 1 [label=7 color="#999933"]\n\t1 -> 1 [label=8 color="#332288"]\n\t1 -> 1 [label=9 color="#cc6677"]\n\t1 -> 1 [label=10 color="#ddcc77"]\n\t1 -> 1 [label=11 color="#117733"]\n\t2 -> 2 [label=0 color="#cc6677"]\n\t2 -> 2 [label=1 color="#ddcc77"]\n\t2 -> 2 [label=2 color="#117733"]\n\t2 -> 2 [label=3 color="#88ccee"]\n\t2 -> 2 [label=4 color="#44aa99"]\n\t2 -> 2 [label=5 color="#882255"]\n\t2 -> 2 [label=6 color="#44aa99"]\n\t2 -> 2 [label=7 color="#999933"]\n\t2 -> 2 [label=8 color="#332288"]\n\t2 -> 2 [label=9 color="#cc6677"]\n\t2 -> 2 [label=10 color="#ddcc77"]\n\t2 -> 2 [label=11 color="#117733"]\n\t3 -> 3 [label=0 color="#cc6677"]\n\t3 -> 3 [label=1 color="#ddcc77"]\n\t3 -> 3 [label=2 color="#117733"]\n\t3 -> 3 [label=3 color="#88ccee"]\n\t3 -> 3 [label=4 color="#44aa99"]\n\t3 -> 3 [label=5 color="#882255"]\n\t3 -> 3 [label=6 color="#44aa99"]\n\t3 -> 3 [label=7 color="#999933"]\n\t3 -> 3 [label=8 color="#332288"]\n\t3 -> 3 [label=9 color="#cc6677"]\n\t3 -> 3 [label=10 color="#ddcc77"]\n\t3 -> 3 [label=11 color="#117733"]\n}\n'  # pylint: disable=line-too-long
```
james-d-mitchell, line 883: You need to escape the quotes in, for example using color=\"#44aa99\" otherwise you are just ending the string, if you see what I mean.

MTWhyte: On second thought, is this definitely the problem? Even though they don't pass, I the test still does run. I think since the outer quotes are different (`'` vs `"`), that Python can parse this ok.

For example in iPython I get

```
In [1]: print('1 + 2 is "3"')
1 + 2 is "3"
```

Possibly it behaves differently in the tests, but if I add a single quote rather than a double at a random place in the string, that does break the syntax and the test doesn't even run.

MTWhyte:
Turns out I misread which test was failing (in short: with my new setup I thought I was just deleting, but it was in fact cutting). Fixed it now!

I was having some trouble shortening the lines and using `+` since the formatter wouldn't be consistent about which quotes to use, so I've just left that as is for the time being.

james-d-mitchell:
Looks good to me, ok to merge @MTWhyte ?

PR #143, opened 2024-04-01
Foundations for v1 release

Joseph-Edwards:
This pull request contains some initial changes made to the bindings in preparation for the release of `libsemigroups` v3. Most of these changes are related to the `KnuthBendix`class and the docs.


james-d-mitchell:
Excellent, I think we expect the ci to fail at this point, amIright?

Joseph-Edwards:
We would expect the ci to fail, since very few of the test files have been updated. However, some of the issues seem to be with installing the requirements, so I'll look into that.

One of the jobs was using Python 3.7 which is too low. We should decide what we want our minimum requirement to be.

PR #147, opened 2024-04-10
Add tests for `Gabow`, `WordGraph` and `Paths`

MTWhyte:
This PR adds some tests for the mentioned stuff, which seemed to get lost in some earlier merging.

### DIFF
--- a/tests/test_paths.py
+++ b/tests/test_paths.py
@@ -1,0 +1,356 @@
+def test_020():
+    "cbegin/end_panislo - 100 node path"
+    wg = WordGraph()
+    n = 100
+    wg.add_nodes(n)
+    wg.add_to_out_degree(2)
+    for i in range(n - 1):
+        wg.add_edge(i, i + 1, i % 2)
+    assert len(list(wg.panilo_iterator(0, 0, POSITIVE_INFINITY))) == 100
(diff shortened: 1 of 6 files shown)


MTWhyte:
I've updated the tests, deleting some stuff that isn't used.

There are a couple of things I've noticed that I'll ask about when convenient, e.g. about how to 'unset' the 'source' node in the `Paths` object.

tests/test_paths.py
```diff
@@ -32,7 +32,6 @@ def test_001():
         w.set_target(i, i % 2, i + 1)
 
     p = Paths(w)
-    assert p.at_end()
```
james-d-mitchell: Just to say that I think this is a bug in libsemigroups, that the return value of `at_end()` is non-deterministic if the Paths object isn't properly initialised (it somehow isn't tracked in the class whether or not it is initialised, but it is implicitly assumed that it is).

james-d-mitchell:
Am I ok to merge this @MTWhyte ?

MTWhyte:
Yeah go ahead!

On 17 Apr 2024, at 16:28, James Mitchell ***@***.***> wrote:

﻿

Am I ok to merge this @MTWhyte<https://github.com/MTWhyte> ?

PR #151, opened 2024-04-12
Add action.hpp

james-d-mitchell:
Adds some support for action.hpp from libsemigroups and there's also some refactoring in knuth-bendix.cpp that shouldn't be part of this PR hence the WIP label.


james-d-mitchell:
@Joseph-Edwards apart from separating out the `Runner` stuff from knuth-bendix.cpp I think this is more or less ready. Things that are missing:
- [ ]  tests
- [ ] doc for `ImageRightAction` 
- [ ]  doc for `RightAction` (the function itself not its methods)

One thing I'm not sure about is: in the doc for `RightAction` the default constructor of `RowActionBMat8` is listed. But the function `RightAction` takes 2 or 3 arguments, and does not have a 0 argument version. How should I go about handling this do you think?

james-d-mitchell:
Ci will fail until I merge 

https://github.com/libsemigroups/libsemigroups/pull/528

into v3

james-d-mitchell:
Note that I also included the generated code for `BMat8` so that I could test the action stuff, but that I didn't really implement anything for BMat8's just yet, will replace that when the time comes.

james-d-mitchell:
This PR builds on #155 and shouldn't be merged until #155 is merged

PR #152, opened 2024-04-12
More additions to the guide

Joseph-Edwards:


james-d-mitchell:
@Joseph-Edwards I don't see any harm in merging this, even though it might not be complete. Still better to have it in the v1 branch than not, what say ye?

Joseph-Edwards:
> @Joseph-Edwards I don't see any harm in merging this, even though it might not be complete. Still better to have it in the v1 branch than not, what say ye?

There's a few things I'd like to add today. And then I can merge whatever I've got by the end of the day?

PR #154, opened 2024-04-15
Fix the doc indentation.

Joseph-Edwards:
This PR adds a script that will fix (most) instances of badly indented docstring, and fixes the current badly indented docstring.


james-d-mitchell:
With the changes to the code generation in `libsemigroups` this script should be redundant eventually but still useful to have it!

PR #155, opened 2024-04-21
Add `Reporter` and `Runner`

james-d-mitchell:


docs/source/conf.py
```diff
@@ -63,11 +63,11 @@
 
 autoclass_content = "both"
 
-# This dictionary should be of the form "bad type" -> "good type", and
-# replacements will be performed globally. Hyperlinks will be added in the
-# signature if "good type" is a valid (potentially user defined) python type
```
Joseph-Edwards, line 68: It'd be good to keep this original comment with the explanation for the hyperlinks

james-d-mitchell: Ah, good catch, I didn't intentionally change this, just got confused when rebasing, will reverse this change.

james-d-mitchell:
I think the CI will fail until https://github.com/libsemigroups/libsemigroups/pull/532 is merged.

docs/source/conf.py
Joseph-Edwards: I think you've changed too much now. The addition of `libsemigroups::BMat8` -> `BMat8` was good, it was just the comment that should've been reverted. Sorry - should've made that more clear!

src/runner.cpp
```diff
@@ -0,0 +1,481 @@
...
+)pbdoc")
+        .value("not_running", Runner::state::not_running, R"pbdoc(
+Indicates that none of :any:`run` , :any:`run_for` , or :any:`run_until`
+has been called since construction or the last call to :any:`init`.
+)pbdoc")
+        .value("dead", Runner::state::dead, R"pbdoc(
+Indicates that none of :any:`run` , :any:`run_for` , or :any:`run_until`
+has been called since construction or the last call to :any:`init`.
+)pbdoc");
+    // TODO repr
```
Joseph-Edwards, line 280: Could a repr be something simple like `f"<Runner instance : {self.current_state()}>"`

Joseph-Edwards (commented):
Other than the addition of a repr/removal of the TODO, this all looks good to me. I noticed a few other TODOs when I started the review, but I think your subsequent pushes in the meantime have removed them.

james-d-mitchell:
> Other than the addition of a repr/removal of the TODO, this all looks good to me. I noticed a few other TODOs when I started the review, but I think your subsequent pushes in the meantime have removed them.

I forgot to add the doc files, which I've done, now, I'm not sure I'm happy with those, can we perhaps discuss at some point?

src/runner.cpp
james-d-mitchell: I could add that, but at the same time, it's not possible to create a `Runner` that isn't also something else, so probably I should delete this TODO instead.

PR #158, opened 2024-04-24
Restructure doc

Joseph-Edwards:
This PR restructures the files in the documentation, and removes a lot of the out of date / non-auto-generated documentation. Moreover, the file structure has been changed to more closely reflect the structure of the table of contents.


Joseph-Edwards:
I'm not sure where to best place `semigroups.rst` and `overlap.rst` hence the draft-ness of this PR. @james-d-mitchell any suggestions?

Joseph-Edwards:
@james-d-mitchell I think this is good now. I think the distinction between algorithms and data structures is a good one, and now the file structure is rational too.

w.r.t merging, are we expecting the macos tests to fail at the moment? Seems like they can't find `libsemigroups` which (hopefully!) is not a result of the changes in this PR.

james-d-mitchell:
It seems that the macosx ci jobs are failing because libsemigroups fails to build, not sure we expect that, or what's going on

Joseph-Edwards:
> It seems that the macosx ci jobs are failing because libsemigroups fails to build, not sure we expect that, or what's going on

Presumably it's something to do with CCache? Is it something you can reproduce locally?

james-d-mitchell:
I think I've fixed this in #159, can you please rebase and see if that helps?

Joseph-Edwards:
> I think I've fixed this in #159, can you please rebase and see if that helps?

It seems to fix it. Thanks @james-d-mitchell!

PR #160, opened 2024-04-25
Update the autodoc directive to improve overloaded functions

Joseph-Edwards:
This PR improves how overloaded functions get rendered in the documentation. Specifically, it formats them in the same that non-overloaded functions get formatted, and adds a level of indentation.

I decided to use this directive to override the standard `autoclass` and `autofunction` directives, since it provides a superset of their functionality. The only downside that I can see is that this causes a `directive already registered` warning. If this is undesirable I can call it something else.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,13 +16,13 @@ install:
 	pip3 install . --use-feature=in-tree-build
 
 black:
-	black setup.py tests/*.py libsemigroups_pybind11/*.py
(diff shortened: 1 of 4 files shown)


Joseph-Edwards:
I've just noticed a few problems with this, specifically with `automodule`. Fixing them now.

james-d-mitchell:
@Joseph-Edwards looks good, but needs to be rebased please

PR #162, opened 2024-05-01
Add only-document-once option

Joseph-Edwards:
This PR introduces the option `:only-document-once:`. Any docstring that starts with this option will only appear once in that functions entry in the documentation, even if that function is overloaded. This is useful when overloaded functions get created as a consequence of types hidden from the user.

In order to specify what the signature of the overloaded function should be (instead of the default `(*args, **kwargs)`), you can specify `:sig=<NEW_SIGNATURE>:` before the `:only-document-once:` option.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -221,9 +221,6 @@ def fix_overloads(app, what, name, obj, options, lines):
     """Indent overloaded function documentation and format signatures"""
 
     for line in lines:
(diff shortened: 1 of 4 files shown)


Joseph-Edwards:
Contrary to my initial thoughts, these changes do in fact in behave nicely with the `.. autosummary::` directive, and therefore no extra changes need to be made in order to get the brief description of a function in the summary column.

PR #165, opened 2024-05-23
Create an environment for developing

Joseph-Edwards:
This PR adds a script that can be sourced to create and activate an environment with the packages and environment variables necessary to develop `libsemigroups_pybind11`.

I had originally hoped that this would be possible using conda, mamba or micromamba. Unfortunately, micromamba doesn't presently support the feature to add environment variables for specific envs, so only conda and mamba are supported.

The contributing guide has also been updated to reflect how a developer should go about setting up their working environment.


james-d-mitchell:
Thanks @Joseph-Edwards !

PR #169, opened 2024-05-28
Add support for AhoCorasick

Joseph-Edwards:
This PR adds the bindings, documentation and tests for the `AhoCorasick` class


Joseph-Edwards:
This will fail until https://github.com/libsemigroups/libsemigroups/pull/545 is merged

Joseph-Edwards:
I've tried to use `iwyu`, but I'm not 100% confident on when I should be using `#include <path>` vs `#include "path"`. My understanding is use `<path>` unless the file is somehow local (which is only `main.hpp` in this case).

Is this the right way to think about it @james-d-mitchell ?

james-d-mitchell:
That's right @Joseph-Edwards enough for our purposes here!

PR #175, opened 2024-07-30
Bipart hpp

james-d-mitchell:
This isn't finished and the CI won't pass until the relevant changes are merged into libsemigroups itself


src/transf.cpp
```diff
@@ -145,6 +145,7 @@ Constructs an uninitialized {} of degree ``0``.
           "copy",
           [](PyClass const& f) { return PyClass(f); },
           fmt::format(
+              // TODO the doc here isn't so good, there's actually no parameter
               R"pbdoc(
 Copy a {0}.
```
Joseph-Edwards, line 150: Should this be `__copy__` to be used with `copy.copy`?

tests/test_bipart.py
```diff
@@ -0,0 +1,99 @@
...
+import pytest
+from element import check_products
+from libsemigroups_pybind11 import (
+    Bipartition,
+    LibsemigroupsError,
+    bipartition,
+    blocks,
+)
+
+# TODO doc the helpers
```
Joseph-Edwards, line 26: This is done right?

src/bipart.cpp
```diff
@@ -0,0 +1,535 @@
...
+              py::arg("blocks"),
+              R"pbdoc(
+Constructs a Blocks object from a list of lists of integers.
+
+This function constructs a :any:`Blocks` object from a list of lists of
+integers, so that the blocks consisting of negative values are
+transverse and those consisting of positive values are not.
+
+:param blocks: the blocks
+:type blocks: list
```
Joseph-Edwards, line 77:
```suggestion
:type blocks: List[List[bool]]
```

src/bipart.cpp
```diff
@@ -0,0 +1,535 @@
...
+*  have the property that if :math:`i`, :math:`i > 0` occurs in *lookup*,
+   then :math:`i - 1` occurs earlier in *lookup*. The value of ``blocks[i]``
+   should represent the index of the block containing ``i``.
+
+For example, if *lookup* is ``[0, 1, 1, 2, 1, 1, 3, 1, 1, 4, 5, 6]`` , then
+the above conditions are satisfied, but if *lookup* is ``[1, 0, 1, 10]`` then
+they are not.
+
+:param lookup: a lookup for the blocks of the bipartition being constructed.
+:type lookup: list
```
Joseph-Edwards, line 303:
```suggestion
:type lookup: List[List[int]]
```

src/bipart.cpp
```diff
@@ -0,0 +1,535 @@
...
+Return a const iterator pointing to the index of the first block.
+
+:complexity:
+   Constant.
+
+:returns:
+   A value of type :any:`Iterator`.
+
+:rtype:
+   Iterator
```
Joseph-Edwards, line 109:
```suggestion
   iterator
```

Joseph-Edwards (changes requested):
Once the function with mangled names have been properly documented (after this is rebased onto the changes from presentation and to_presentation), this looks good to me!

tests/test_bipart.py
james-d-mitchell: Yup, done

src/transf.cpp
james-d-mitchell: Also, I think I was going for having both `copy` and `__copy__` but haven't done the second here yet, I think I'll do that in a separate PR, there are some more things that need to be fixed for transformations too.

src/bipart.cpp
james-d-mitchell: Are you sure about this? If we use `Iterator` it links to the relevant part of the python docs.

src/old/sims.cpp
```diff
@@ -34,7 +34,10 @@
 
 // libsemigroups....
 #include <libsemigroups/presentation.hpp>  // for Presentation
+<<<<<<< HEAD:src/old/sims.cpp
```
james-d-mitchell, line 37:
```suggestion
```

src/old/sims.cpp
```diff
@@ -34,7 +34,10 @@
...
 #include <libsemigroups/sims.hpp>          // for Sims1
+=======
+>>>>>>> 03328ae (Add Blocks):src/sims1.cpp
```
james-d-mitchell, line 40:
```suggestion
```

src/bipart.cpp
Joseph-Edwards: That link isn't being created for me, but nor is it being created when I use `iterator`. Since `iterator` isn't a type in python, I suppose it doesn't really matter. The actual return type seems to be `pybind11_builtins.iterator`.

james-d-mitchell: Ok, I'll go with `iterator` thanks!

Joseph-Edwards (approved):

PR #180, opened 2024-08-05
Update forest.hpp for v1/3

james-d-mitchell:


src/forest.cpp
```diff
@@ -37,122 +37,230 @@ namespace py = pybind11;
...
+        py::arg("i"),
+        R"pbdoc(
+Returns a list containing the labels of the edges on the path from a root node
+to *i*.
+
+:param i: the node.
+:type i: int
+
+:raises LibsemigroupsError:
+  if ``i`` is greater than or equal to :any:`number_of_nodes`.
```
Joseph-Edwards, line 240:
```suggestion
  if *i* is greater than or equal to :any:`number_of_nodes`.
```

Joseph-Edwards (changes requested):
I think some of the docs need to be reordered so each function goes:
- Description
- Params
- Param type
- Return
- Return type
- Exceptions
- Complexity

(Or some sublist of this)

I've noticed that the generate-pybind script swaps them around sometimes for some reason. Only just spotted this, so I may have missed this in previous reviews - sorry!

Other than that, and the minor `param` -> *param* changes I've suggested below, this looks good to me

james-d-mitchell:
@Joseph-Edwards I think this is good to go.

Joseph-Edwards (approved):

PR #181, opened 2024-08-06
Update gabow.hpp for v1/3

james-d-mitchell:


Joseph-Edwards (changes requested):
It seems that the documentation introduces a lot of sphinx warnings, but these almost all pertain to `WordGraph` or `WordGraph::some_function`, so they will hopefully go when we do that PR.

The one warning that we should probably address now is
```
WARNING: more than one target found for 'any' cross-reference 'word_graph': could be :py:meth:`_libsemigroups_pybind11.RightActionPPerm16List.word_graph` or :py:meth:`_libsemigroups_pybind11.Gabow.word_graph`
```

This is probably another need for the pseudo-namespace stuff again, if possible. Other than that, everything looks good to me. Thanks @james-d-mitchell !

Joseph-Edwards (approved):

PR #183, opened 2024-08-10
Fix return vals

james-d-mitchell:
Previously the lambdas used for `Paths` mem fns were returning by value in some cases, leading to confusing weirdness.


libsemigroups_pybind11/paths.py
```diff
@@ -6,11 +6,28 @@
...
+
+def count(p: Paths | ReversiblePaths) -> int | type(POSITIVE_INFINITY):
+    result = p._count()
+    if result == POSITIVE_INFINITY:
+        return POSITIVE_INFINITY
+    return result
+
+
+Paths.count = count
+doc = Paths._count.__doc__
```
james-d-mitchell, line 27: @Joseph-Edwards Is there a better way of achieving this? Basically, I'm removing the signature of the function `Paths._count` from its doc so that I can copy the rest of it to `Paths.count.__doc__`.

Joseph-Edwards: If `_count` is never going to be used, you could remove its documentation from the .cpp file and enter it in the .py file without the signature. Then you should just be able to copy it without removing the signature.

I think there may also be a pybind11 option that stops the inclusion of the signature in the doc, but that can only be set on a module level, so would have more unintended consequences than helpful consequences.

From the autodoc web page:
> If you document decorated functions or methods, keep in mind that autodoc retrieves its docstrings by importing the module and inspecting the __doc__ attribute of the given function or method. That means that if a decorator replaces the decorated function with another, it must copy the original __doc__ to the new function.

So I don't think there's much else we can do.

james-d-mitchell: Ah, good idea to just move the doc, I'll do that (not now). Thanks @Joseph-Edwards !

PR #186, opened 2024-08-23
Update words for v1/v3

Joseph-Edwards:
This PR updates the functionality of words in prep for the release of v1. I would like to add a few more test cases, but otherwise this is good to go.


Joseph-Edwards:
CI will fail until https://github.com/libsemigroups/libsemigroups/pull/584 is merged

src/words_old.cpp
james-d-mitchell, line 1: Please delete this file

PR #187, opened 2024-09-09
Update Order for v1/v3

Joseph-Edwards:
This PR adds some comparison functions, and documentation for the Order class.

### DIFF
--- a/src/order.cpp
+++ b/src/order.cpp
@@ -97,8 +97,6 @@ Compare two values of type :any:`str` or ``List[int]`` using using lexicographic
 :returns: The boolean value ``True`` if *x* is lexicographically less than *y*, and ``False`` otherwise.
 :rtype: bool
 
-:exceptions: This function guarantees not to throw a ``LibsemigroupsError``.
-
 :complexity: At most :math:`O(n)` where :math:`n` is the minimum of the length of *x* and the length of *y*.
 )pbdoc");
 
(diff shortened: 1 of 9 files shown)


src/order.cpp
```diff
@@ -45,14 +35,172 @@ namespace py = pybind11;
...
+:param x: the first object for comparison. 
+:type x: str | List[int]
+
+:param y: the second object for comparison.
+:type y: str | List[int]
+
+:returns: The boolean value ``True`` if *x* is less than *y* with respect to the recursive path ordering, and ``False`` otherwise.
+:rtype: bool
+
+:exceptions: This function is and is guaranteed never to throw.
```
james-d-mitchell, line 189: This line is sus

src/order.cpp
```diff
@@ -45,14 +35,172 @@ namespace py = pybind11;
...
+:type x: str | List[int]
+
+:param y: the second object for comparison.
+:type y: str | List[int]
+
+:returns: The boolean value ``True`` if *x`* is short-lex less than *y*, and ``False`` otherwise.
+:rtype: bool
+
+:exceptions: This function guarantees not to throw a ``LibsemigroupsError``.
+
```
james-d-mitchell, line 139: I think we weren't including these lines in the python doc

src/order.cpp
```diff
@@ -45,14 +35,172 @@ namespace py = pybind11;
...
+:param x: the first object for comparison. 
+:type x: str | List[int]
+
+:param y: the second object for comparison.
+:type y: str | List[int]
+
+:returns: The boolean value ``True`` if *x* is lexicographically less than *y*, and ``False`` otherwise.
+:rtype: bool
+
+:exceptions: This function guarantees not to throw a ``LibsemigroupsError``.
```
james-d-mitchell, line 100: Ditto

james-d-mitchell (changes requested):
No couple of quibbles, then merge

src/order.cpp
Joseph-Edwards: Ah my bad. I'll change it in this PR and open an issue to get rid of it elsewhere (this isn't the first time I've included them).

PR #189, opened 2024-09-10
Word graph hpp

james-d-mitchell:
Just need to finish the doc and resolve a few issues, then this is good to go.


james-d-mitchell:
This will fail until the corresponding PR in libsemigroups is merged.

etc/replace-strings-in-doc.py
```diff
@@ -53,6 +53,7 @@ def dive(path):
     "ImageLeftActionPPerm16PPerm16": "ImageLeftAction",
     "ImageRightActionPPerm16PPerm16": "ImageRightAction",
     "aho_corasick_dot.*": "",
+    "word_graph_dot.*": "Returns a Dot object representing a word graph.",
```
Joseph-Edwards, line 56: If #190 is merged, there may be a slight merge conflict here. The solution would be to delete the `"aho_corasick_dot.*": "",` line, and move `"word_graph_dot.*": "Returns a Dot object representing a word graph.",`  from this dictionary to the new dictionary called `docstring_replacements` in `conf.py`. Sorry for creating more work!

Joseph-Edwards:
Unfortunately, there is no version of NumPy that supports both Python 3.8 and Python 3.12. Unless you object @james-d-mitchell, I'm going to open a PR to drop support for Python 3.8 (its end-of-life next month anyway).

I've also opened Issue #192 which highlights a standard we could decide to conform to relating to which versions of Python (and other dependencies) we should aim to support. I'd be interested to hear any thoughts about it.

james-d-mitchell:
Dropping 3.8 sounds good

Joseph-Edwards (approved):

PR #191, opened 2024-09-16
Add pbr for v1

Joseph-Edwards:
This updates pbr in preparation for v1. Since, in `libsemigroups` there is no "checks" version of multiplying `PBR`s, you can't really do very much with these objects, but they're here nonetheless!


Joseph-Edwards:
The doc will have some bad signatures until #190 is merged.

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+:raises LibsemigroupsError: if the resultant PBR:
+
+    *  would not describe a binary relation on an even number of points; or
+    *  would have a point related to a point that is greater than degree();
+)pbdoc");
+    thing.def(py::init([](PBR::vector_type<uint32_t> x) { return to_pbr(x); }),
+              py::arg("x"),
+              R"pbdoc(
+Construct from adjacencies 0 to :math:`2n - 1`.
```
james-d-mitchell, line 138: 0 should be inside double backticks here

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+and the vector in position :math:`i` is the list of points adjacent to :math:`i`
+in the :any:`PBR` constructed.
+
+:param x: the container of vectors of adjacencies.
+:type x: List[List[int]]
+
+:raises LibsemigroupsError: if the resultant PBR:
+
+    *  would not describe a binary relation on an even number of points; or
+    *  would have a point related to a point that is greater than degree();
```
james-d-mitchell, line 151: Reference degree?

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+Returns the number of points of a :any:`PBR`.
+
+:returns: A value of type ``int``.
+:rtype: int
+
+:complexity: Constant.
+)pbdoc");
+    thing.def(
+        "__getitem__",
+        [](const PBR& a, size_t b) { return a[b]; },
```
james-d-mitchell, line 181: Is this safe if b is out of bounds?

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+        py::arg("x"),
+        R"pbdoc(
+Returns the identity PBR with degree ``x.degree()``.
+
+This member function returns a new :any:`PBR` with degree equal to the degree of
+*x* , where every value is adjacent to its negative. Equivalently, :math:`i` is
+adjacent :math:`i + n` and vice versa for every :math:`i` less than the degree
+:math:`n`.
+
+:param x: The identity.
```
james-d-mitchell, line 208: This seems sus

tests/test_pbr.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+This module contains some tests for PBRs.
+"""
+
+# pylint: disable= missing-function-docstring, invalid-name
+
+from copy import copy
+import pytest
+from libsemigroups_pybind11 import PBR, pbr
+
```
james-d-mitchell, line 18: Maybe add some me tests for the left right constructor?

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+    *  would have a point related to a point that is greater than degree();
+    *  *x* contains a list of points related to a point that is not sorted.
+
+)pbdoc");
+    thing.def("degree",
+              &PBR::degree,
+              R"pbdoc(
+Returns the degree of a :any:`PBR`.
+
+Returns the degree of a :any:`PBR` , where the *degree* of a :any:`PBR` is half
```
james-d-mitchell, line 160: Extra space before comma

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+:type n: int
+)pbdoc");
+    thing.def(py::init([](PBR::vector_type<int32_t> left,
+                          PBR::vector_type<int32_t> right) {
+                return to_pbr(left, right);
+              }),
+              py::arg("left"),
+              py::arg("right"),
+              R"pbdoc(
+Construct from adjacencies 1 to :math:`n` and -1 to :math:`-n`.
```
james-d-mitchell, line 115: Mixture of math, plain numbers and double back ticks

james-d-mitchell (changes requested):

james-d-mitchell:
Can you maybe add the checked multiplication, and the other comparison operators? Shouldn't be much code

src/pbr.cpp
```diff
@@ -0,0 +1,233 @@
...
+    thing.def(
+        "copy",
+        [](PBR const& self) { return PBR(self); },
+        R"pbdoc(
+Copy a PBR object.
+
+:returns: A copy.
+:rtype: PBR
+)pbdoc");
+    thing.def(py::init<size_t>(),
```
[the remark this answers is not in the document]
Joseph-Edwards: You could call `product_inplace` and store the results in it? If that's worth anything?

Joseph-Edwards:
> Can you maybe add the checked multiplication, and the other comparison operators? Shouldn't be much code

Is this simply the case of checking the degrees are the same? Or is there more to it?

src/pbr.cpp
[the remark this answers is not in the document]
james-d-mitchell: Ah, ok, sounds good

james-d-mitchell:
> > Can you maybe add the checked multiplication, and the other comparison operators? Shouldn't be much code
> 
> Is this simply the case of checking the degrees are the same? Or is there more to it?

Just checking the degrees are equal and also validating the PBRs, that should do it.

PR #193, opened 2024-09-20
Update obvinf.hpp for v1/3

james-d-mitchell:


src/obvinf.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "main.hpp"  // for init_obvinf
+
+namespace py = pybind11;
+
+namespace libsemigroups {
+
+  void init_obvinf(py::module& m) {
+    m.def(
+        "is_obviously_infinite",
+        [](Presentation<word_type> const& p) {
```
Joseph-Edwards, line 38: Do we want to do this for `Presentation<std::string>` too?

james-d-mitchell: Good point forgot that one!

src/knuth-bendix.cpp
Joseph-Edwards, line 1: Should this function be moved into `obvinf.cpp` to be with the rest of them?

src/knuth-bendix.cpp
```diff
@@ -550,15 +550,32 @@ is returned.
...
+This function returns ``True`` if the quotient of the finitely presented
+semigroup or monoid defined by the :any:`KnuthBendix` object *kb* is obviously
+infinite; ``False`` is returned if it is not.
+
+:param kb: the :any:`KnuthBendix` instance.
+:type kb: KnuthBendix
+
+:returns:
+    Whether or not the quotient defined by a :any:`KnuthBendix` instance is
+    obviously infinite.
```
Joseph-Edwards, line 573:
```suggestion
:returns:
    Whether or not the quotient defined by a :any:`KnuthBendix` instance is
    obviously infinite.
:rtype:
    bool
```

Joseph-Edwards (changes requested):
Happy to merge this modulo the few minor suggestions. Thanks @james-d-mitchell!

Joseph-Edwards (approved):

PR #195, opened 2024-10-02
Add Ukkonen for v1

Joseph-Edwards:
This PR updates `Ukkonen` for the release of v1.


src/main.cpp
```diff
@@ -161,11 +161,12 @@ namespace libsemigroups {
     init_dot(m);
     init_matrix(m);
     init_froidure_pin_base(m);
+    init_ukkonen(m);
 
 #ifdef VERSION_INFO
     m.attr("__version__") = VERSION_INFO;
 #else
-    m.attr("__version__") = "dev";
+    m.attr("__version__")                 = "dev";
```
james-d-mitchell, line 169: ??

james-d-mitchell (approved):
Obviously have to merge the libsemigroups PR first, but this looks great otherwise

src/ukkonen.cpp
```diff
@@ -0,0 +1,936 @@
...
+
+Find the index of a word in the suffix tree.
+
+If the *w* is one of the words that the suffix tree contains
+(the words added to the suffix tree via :any:`ukkonen.add_word`), then this function
+returns the index of that word. If the word *w* is not one of the words that the
+suffix tree represents, then :any:`UNDEFINED` is returned.
+
+:param w: the word to check.
+:type w: str|List[int]
```
james-d-mitchell, line 60: No idea if this impacts the generated doc or not but isn't there some missing whitespace here?

src/ukkonen.cpp
```diff
@@ -0,0 +1,936 @@
...
+:sig=(self: Ukkonen, w: str|List[int]) -> None:
+:only-document-once:
+Throw if the word *w* contains a letter equal to any of the unique letters added
+to the end of words in the suffix tree.
+
+This function throws an exception if the word *w* contains a letter equal to any
+of the unique letters added to the end of words in the suffix tree.
+
+:param w: the word to check.
+:type w: str|List[int]
```
james-d-mitchell, line 85: Missing whitespace

src/main.cpp
Joseph-Edwards: This seems odd to me too, but I've double checked and this is indeed what `clang-format-15` wants. Should we add one of the "Don't format me please things"?

src/ukkonen.cpp
Joseph-Edwards: Between `str` and `list`?

james-d-mitchell: Yup

src/ukkonen.cpp
```diff
@@ -0,0 +1,936 @@
...
+
+:param l: the left most index and value of the data member l (defaults to ``0``). 
+:type l: int
+
+:param r: one after the right most index and value of the data member r (defaults to ``0``). 
+:type r: int
+
+:param parent: parent of the node being constructed (defaults to :any:`UNDEFINED`).
+:type parent: int
+)pbdoc");
```
Joseph-Edwards, line 640: I'm having some issues with this definition when compiling with `clang`. The error I get is:
```cpp
src/ukkonen.cpp:628:32: error: use of overloaded operator '=' is ambiguous (with operand types 'py::arg' and 'const Undefined' (aka 'const Constant<0, detail::Max>'))
  628 |              py::arg("parent") = UNDEFINED,
      |              ~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~
```

I can make this go away by casting `UNDEFINED` to a `node_index_type` but this feels a bit dubious. What do you think @james-d-mitchell?

james-d-mitchell: I think casting to a `node_index_type` is safe here, it'll be passed as such to the underlying C++ constructor anyway, so might as well be statically cast here.

PR #199, opened 2024-11-06
Handle returning `POSITIVE_INFINITY` in `Paths`

Joseph-Edwards:
This PR updates the `Paths` class. Specifically, it defines some decorators, one of which allows a function to return `POSITIVE_INFINITY`. Furthermore, it documents the classes of each of the constants we define, so that more types in function signatures get converted to URLs.


james-d-mitchell:
Excellent thanks 🙏 @Joseph-Edwards

PR #200, opened 2024-11-08
Add `SchreierSims` for v1

Joseph-Edwards:
This PR adds the `SchreierSims` class and associated helpers in advance of the v1 release. Supersedes #109.

## TODO

- [x] Add the bindings to the `cpp` file;
- [x] Expose the functionality in `libsemigoups_pybind11`;
- [x] Add the the pages to the doc;
- [x] Wrap the class in a Python class (after #197 is merged);
- [x] Format the documentation and the signatures;
- [x] Add tests.

### DIFF
--- a/docs/source/main-algorithms/schreier-sims/index.rst
+++ b/docs/source/main-algorithms/schreier-sims/index.rst
@@ -8,7 +8,7 @@ Schreier-Sims
 =============
 
 This page describes the functionality related to the Schreier-Sims algorithm for
-computing a base and strong generating set of a permutation group in
+computing a stabilizer chain of a permutation group in
 ``libsemigroups_pybind11``.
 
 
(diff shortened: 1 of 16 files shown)


Joseph-Edwards:
The Mac tests are failing because `SchreierSims` and `FroidurePin` have identical constructors. @james-d-mitchell, do you think we should edit these classes to inherit from a common `InitFromGenerators` abstract base class, or just suppress the warning?

james-d-mitchell:
> The macOS tests are failing because `SchreierSims` and `FroidurePin` have identical constructors. @james-d-mitchell, do you think we should edit these classes to inherit from a common `InitFromGenerators` abstract base class, or just suppress the warning?

Maybe simplest to just add a TODO to do that later and to disable the linting error for now

docs/source/main-algorithms/schreier-sims/index.rst
```diff
@@ -0,0 +1,19 @@
...
+
+   Distributed under the terms of the GPL license version 3.
+
+   The full license is in the file LICENSE, distributed with this software.
+
+Schreier-Sims
+=============
+
+This page describes the functionality related to the Schreier-Sims algorithm for
+computing a base and strong generating set of a permutation group in
```
james-d-mitchell, line 11:
```suggestion
computing a stabilizer chain of a permutation group in
```

Issue #203, opened 2024-11-21
Add a Python 3.13 job to the CI


Joseph-Edwards:
Since Python 3.13 is now a stable release, we should probably check we are compatible with it.

PR #204, opened 2024-11-21
Bump Python version

Joseph-Edwards:
Addresses #203 by adding Python 3.13 to the CI.


james-d-mitchell:
Thanks for this @Joseph-Edwards. Is this a standard approach to deal with different dependencies on different versions of python? If not, then I think it might be better to just drop support for python 3.9, and have a fixed set of requirements. Otherwise, we're just storing up problems for the future I'm afraid.

Joseph-Edwards:
> Thanks for this @Joseph-Edwards. Is this a standard approach to deal with different dependencies on different versions of python? If not, then I think it might be better to just drop support for python 3.9, and have a fixed set of requirements. Otherwise, we're just storing up problems for the future I'm afraid.

I'm not sure if this is standard. I upgraded the version of Python in my local `libsemigroups _pybind11_dev` environment to 3.13, so I needed to see which versions of the dependencies I should install locally. It made sense to try and install versions that also happened to work with Python 3.9 -- 3.12  which is why I knocked up the quick script. I only intended this script to be run every time the list of Python versions we support is updated. 

As I see it, we have two different types of requirements that server two different purposes; package requirements and developer requirements.

## Package requirements
These are the packages required for our project to work, and are specified in `setup.py` and `pyproject.toml` (to conform with [PEP 518](https://peps.python.org/pep-0518/) we may have to edit our packaging process, but that's an issue for a later time). Since we want our package to be used by as many people as possible, these requirements should be as relaxed as possible.

## Developer requirements
By developer requirements, I mean the packages any developer working on our Python bindings should have installed. These are specified in the `requirements.txt` file, and are also used by the CI. To me, it makes sense for these to as strict as possible (i.e. all requirements have the form `pkg_name==version`), to maximise reproducibility between developers and the CI. It is these requirements that the Python script in this PR aims to resolve.

I hope this clears things up @james-d-mitchell, but if not, would you mind explaining why dropping support for Python 3.9 would resolve the issue. Thanks

james-d-mitchell:
Thanks @Joseph-Edwards that does indeed clear it up, I'm happy to merge. Thanks for this and the detailed explanation

PR #211, opened 2024-12-17
Minor doc fixes

Joseph-Edwards:
This PR makes some minor changes to the doc. Primarily, it fixes the documentation for some overloaded constructors and `init` functions, and resolves from broken links.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+setuptools==75.6.0
 graphviz==0.20.3
 numpy==2.0.2
 packaging==24.2
--- a/src/aho-corasick.cpp
+++ b/src/aho-corasick.cpp
@@ -371,7 +371,7 @@ index *start*, and traversing using the letters in the word *w*.
 :param ac: object to traverse.
 :type ac: AhoCorasick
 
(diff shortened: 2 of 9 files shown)


Joseph-Edwards:
It's not immediately clear to me why the older versions of Python have stopped working. My only guess is that the `ubuntu-latest` and `macos-latest` GitHub Action virtual machines may have been updated recently? I'll look into it and try and resolve.

Joseph-Edwards:
I messed around with the CI a bit on my fork and, to my surprise, changing the runner version didn't resolve the issue. I noticed that `setuptools` was only being installed alongside Python Versions 3.12 and 3.13, which may has something to do with [PEP 632](https://peps.python.org/pep-0632/) and `setuptools` being preferred  over (the now deprecated) `distutils`. Adding `setuptools` to `requirements.txt` seems to resolve the issues.

I really have no idea why the tests were passing the other day, but not today. My new best guess is that one of our dependencies received an update which dropped `distutils` in support of `setuptools`. This is probably something we should be wary of when it comes to packaging and making a release.

PR #216, opened 2025-02-19
Add `to` function for `FroidurePin`

Joseph-Edwards:
This PR adds a `to` function, and the implementation to make it work when converting to a `FroidurePin` type. This is to prove this is possible; other types will be added in due course.


src/froidure-pin.cpp
```diff
@@ -739,14 +740,16 @@ elements are sorted, or :any:`UNDEFINED` if *i* is greater than
...
+    // TODO(0) Something for Congruence?
+    bind_froidure_pin<
+        detail::KBE<detail::KnuthBendixImpl<detail::RewriteFromLeft>>>(
+        m, "KBERewriteFromLeft");
+    bind_froidure_pin<
+        detail::KBE<detail::KnuthBendixImpl<detail::RewriteTrie>>>(
+        m, "KBERewriteTrie");
+    bind_froidure_pin<detail::TCE>(m, "TCE");
+    bind_froidure_pin<detail::KE<std::string>>(m, "KEString");
+    bind_froidure_pin<detail::KE<word_type>>(m, "KEWord");
```
Joseph-Edwards, line 752: Are these names okay? Happy to receive suggestions.

james-d-mitchell: I don't think these are required at all

Joseph-Edwards: Without them we can't convert the return type of `to` into a Python type.

james-d-mitchell: Ah wait I misunderstood the purpose of this, yes, I'm happy with the names! Sorry

Joseph-Edwards:
I just need to finish splitting the documentation into the correct locations then this will be good to go.

libsemigroups_pybind11/to.py
```diff
@@ -0,0 +1,156 @@
...
+from .congruence import Congruence
+from .froidure_pin import FroidurePin
+from .knuth_bendix import KnuthBendix
+from .presentation import Presentation, InversePresentation
+from .todd_coxeter import ToddCoxeter
+from .detail.cxx_wrapper import to_cxx
+
+
+def _nice_name(type_list):
+    """Convert an itterable of type-like things into a string"""
```
james-d-mitchell, line 43:
```suggestion
    """Convert an iterable of type-like things into a string"""
```

james-d-mitchell:
One spelling mistake to fix, then I'm happy for this to be merged. 

One further (not important) comment: I don't completely understand why some of the changes that are already in `v1` are showing as being new in this PR, but maybe the rebasing was difficult?

Joseph-Edwards:
> One spelling mistake to fix, then I'm happy for this to be merged. 
> 
> 
> 
> One further (not important) comment: I don't completely understand why some of the changes that are already in `v1` are showing as being new in this PR, but maybe the rebasing was difficult? 

There was quite a lengthy rebase. I thought I was thorough, but I must've missed something.

The one thing that remains to be done is to fix the `FroidurePin` constructor to work with different element types. In its current state, I don't think objects constructed from other cxx type `FroidurePin` objects can return their elements. What should do about this in this PR?

james-d-mitchell:
Let’s discuss tomorrow am. I think this is probably complete, but I don’t exactly understand your last comment.

Joseph-Edwards:
I think this is good to go

libsemigroups_pybind11/froidure_pin.py
```diff
@@ -218,6 +236,7 @@ def sorted_at(self: Self, i: int) -> Element:
         return self._cxx_obj.sorted_at(i)
 
     @may_return_undefined
+    @_returns_element
     def current_position(self: Self, x: Element) -> Element:
         return self._cxx_obj.current_position(to_cxx(x))
```
james-d-mitchell, line 241: This returns an int not an element iirc

james-d-mitchell (commented):
Looks good couple of minor things then let’s merge

james-d-mitchell:
Also the CI should run without any errors :)

PR #198, opened 2024-11-04
Add support for Sims for v1 release

reiniscirpons:
This is a draft PR adds python bindings of the `Sims` class from `libsemigroups` for the v1 release. Mainly put out here to increase the bus factor.

See also https://github.com/libsemigroups/libsemigroups/pull/607 for the corresponding PR in `libsemigroups`.

PR #220, opened 2025-03-05
Add support for Sims for v1 release, version 2.0, now with better branch name!

reiniscirpons:
Same as #198 but I changed the branch name because I was dumb in the past and didnt anticipate working on more than one feature for `v1`.


james-d-mitchell:
Missing asserts I think @Joseph-Edwards

james-d-mitchell:
Yay!

PR #227, opened 2025-03-07
Fix presentation_examples tests

james-d-mitchell:


james-d-mitchell:
CI will fail here until https://github.com/libsemigroups/libsemigroups/pull/675 is merged

PR #235, opened 2025-04-01
Use proper classes

james-d-mitchell:
This PR intends to replace the "functions pretending to be classes" in `libsemigroups_pybind11` with proper classes derived from `CxxWrapper`. 

The functions that need to be converted are:

- [X] `Congruence`
- [x] `KnuthBendix`
- [x] `Kambites`
- [x] `ToddCoxeter`
- [x] `Presentation`
- [x] `InversePresentation`
- [x] `FroidurePin`
- [x] `Konieczny`

and I think that's all. @reiniscirpons, I've implemented something that ticks all the boxes for me for the `Congruence` class, if you have any feedback, then I'd be happy to hear it!


reiniscirpons:
Had a look, it works on my system and I can now do things like `isinstance(k, Kambites)` and subclass on `Kambites` and use `Kambites` as a type hint. However, member functions that return a reference to the cxx object now cause issues, e.g.
```python3
k1 = Kambites(Word=str)
assert isinstance(k1, Kambites) # Passes
k2 = k1.init()
assert isinstance(k2, Kambites) # Fails
```
since the `init` function returns an instance of the underlying `KambitesString` or `KambitesWord` class not the `Kambites` wrapper class. In the case of Kambites one fix would just be to manually wrap these functions with an extra `Kambites` conversion at the end. Alternatively we could just write some function converting the internal instances to the appropriate wrapped instances and modify the cxxwrapper to automatically wrap them.

E.g. something like 
```python3
_internal_to_wrapper_dict = {
    KambitesWords: Kambites,
    KambitesStrings: Kambites,
    PresentationWords: Presentation,
    PresentationStrings: Presentation, # Every time we add a wrapper register it here
}

def internal_to_wrapper(x):
    if type(x) in _internal_to_wrapper_dict:
        return _internal_to_wrapper_dict[type(x)](x)
    return x
```
and then we wrap the return of `__getattr__` in `CxxWrapper` with `internal_to_wrapper`. Though I guess we might need to do this recursively if the return type is a list or tuple. This does feel like a hacky solution though.

Another way would be to somehow make `KambitesWords` and `KambitesStrings` somehow be recognized as subclasses of `Kambites`. But I have no idea how to do this or if it might break everything if we try.

james-d-mitchell:
Thanks for the comments @reiniscirpons, that's an oversight on my behalf about the return value of `init` et al.

Some comments: I don't think the new approach really uses `__getattr__` in `CxxWrapper` at all, for a couple of reasons (mostly pointed out by @Joseph-Edwards):
1. if we don't define a method with name `foo` in, say, `Kambites`, then tab completion doesn't work in ipython. This is arguably not super important, but it's nice if it does work;
2. no documentation is generated for `foo`. This is critical, for example, in `Congruence` and `Kambites` some methods are declared explicitly in the class definition (`has` and `get` and `number_of_classes` in `Congruence`), and others are implicit and "installed" via `_copy_cxx_mem_fns`. If we don't do `_copy_cxx_mem_fns`, then there'd be no doc for the methods installed this way in `Congruence`. On the other hand, if we don't `_copy_cxx_mem_fns`, then we could get the doc from `CongruenceWords` in `_libsemigroups_pybind11` but then we won't have the doc for the explicit methods such as `get` and `has`.

All of that is to say that I think the best thing is just to explicitly declare the methods that should return `self`, and add a decorator to them, as we've done with `number_of_classes` for example. Sound ok?

reiniscirpons:
> I don't think the new approach really uses `__getattr__` in `CxxWrapper` at all

Ah ok, I overlooked the `copy_cxx_mem_fns` function and misunderstood how the wrapping works. 

> All of that is to say that I think the best thing is just to explicitly declare the methods that should return self, and add a decorator to them, as we've done with `number_of_classes` for example. Sound ok?

Sounds good, this is probably the simplest fix. One extra issue might be if we have any member function which returns a different wrapped type (e.g. a Congruence returning a ToddCoxeter), but for Congruence this is already handled a with the `get` function and I imagine all other such conversion go through the `to` function, so should be fine.

james-d-mitchell:
I think I've accounted for the issues that you pointed out @reiniscirpons if you have any other comments, please let me know!

Joseph-Edwards:
In 6019bc3, we update our custom extension to look for inserted signatures at the start of a docstring. If the name of the function in the inserted signature does not match the name of the function being documented (e.g. `knuth_bendix_non_trivial_classes` vs `non_trivial_classes`) then we remove those inserted lines from the doc.

This means that we can remove the  `options.disable_function_signatures()` stuff from `FroidurePin` if that is desired.

james-d-mitchell:
I think this is now complete from my perspective @Joseph-Edwards

Joseph-Edwards:
> I think this is now complete from my perspective @Joseph-Edwards 

Sweet! Thanks @james-d-mitchell. I'll try and review the changes/use the package over the next couple of days and try and get it merged.

james-d-mitchell:
Oooh right I merge that but didn't check, I'll fix that now

docs/source/data-structures/adapters/imageleftaction.rst
```diff
@@ -0,0 +1,31 @@
+.. Copyright (c) 2024 James Mitchell
+
+   Distributed under the terms of the GPL license version 3.
+
+   The full license is in the file LICENSE, distributed with this software.
+
+.. currentmodule:: libsemigroups_pybind11
+
+The ImageLeftAction class
+==========================
```
james-d-mitchell, line 10:
```suggestion
=========================
```

docs/source/data-structures/presentations/to-inverse-present.rst
```diff
@@ -32,25 +32,25 @@ options are possible:
 
 .. _presentation-to-inverse-presentation:
 
-Converting a :any:`PresentationStrings` to an :any:`InversePresentationStrings`
+Converting a :any:`Presentation` to an :any:`InversePresentation`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
james-d-mitchell, line 36:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

libsemigroups_pybind11/action.py
```diff
@@ -14,359 +14,349 @@
...
+            * **seeds** (*List[Point]*) -- at least one seed point for the action.
+            * **func** (*Callable[[Point, Element], Point]*) -- the function defining the action.
+            * **side** (:py:class:`side <side>`)-- the side of the action.
+
+        :raises TypeError:
+            if *generators* or *seeds* is not a list.
+        :raises ValueError:
+            if *generators* or *seeds* has length ``0``.
+        :raises KeyError:
+            if the action defined by the arguments is not defined.
```
james-d-mitchell, line 200: This could be better

docs/source/main-algorithms/froidure-pin/class.rst
```diff
@@ -0,0 +1,77 @@
...
+   FroidurePin.suffix
+   FroidurePin.to_sorted_position
+
+Full API
+--------
+
+.. autoclass:: FroidurePin
+   :members:
+   :class-doc-from: init
+   :special-members: __init__
```
Joseph-Edwards, line 72:
```suggestion
```
Since  we have `:class-doc-from: init`, we don't need `:special-members: __init__`.

docs/source/main-algorithms/froidure-pin/helpers.rst
```diff
@@ -29,5 +29,9 @@ Contents
 Full API
 --------
 
+.. currentmodule:: libsemigroups_pybind11.froidure_pin
+
 .. automodule:: libsemigroups_pybind11.froidure_pin
    :members:
+   :imported-members:
+   :exclude-members: FroidurePin, FroidurePinBase
```
Joseph-Edwards, line 37:
```suggestion
   :exclude-members: FroidurePin
```

docs/source/main-algorithms/froidure-pin/to-froidure-pin.rst
```diff
@@ -6,11 +6,11 @@
 
 .. currentmodule:: libsemigroups_pybind11
 
-Converting to a FroidurePinPBR
+Converting to a FroidurePin
 ==============================
```
Joseph-Edwards, line 10:
```suggestion
Converting to a FroidurePin
===========================
```

libsemigroups_pybind11/action.py
```diff
@@ -6,367 +6,356 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 
-# pylint:disable=no-name-in-module, unused-import, protected-access,
-# pylint:disable=missing-function-docstring
+# pylint:disable=no-name-in-module, unused-import,
+# pylint:disable=missing-function-docstring,
```
Joseph-Edwards, line 10: There are a few unused imports in this file that don't get warned about due to these pylint suppressions. Since what we are now doing is more Pythonic than it was before, I think it's probably wise to remove the `pylint:disable` stuff and heed their warnings.

(This is the case for all of the python files in this PR, but I'll only write this comment once)

libsemigroups_pybind11/action.py
```diff
@@ -6,367 +6,356 @@
...
+for _type in Action._py_template_params_to_cxx_type.values():  # pylint: disable=protected-access
+    _register_cxx_wrapped_type(_type, Action)
 
-        if self._cxx_obj is not None:
-            self._cxx_obj.add_generator(to_cxx(x))
-        else:
-            self._generators.append(x)
-        return self
+########################################################################
+# RightAction python class TODO
```
Joseph-Edwards, line 282:
```suggestion
# RightAction python class
```

libsemigroups_pybind11/action.py
```diff
@@ -6,367 +6,356 @@
...
-    """
-    # TODO probably this will generate unhelpful error messages
-    return Action(
-        Func=Func,
-        Element=kwargs["Element"],
-        Point=kwargs["Point"],
-        Side=side.right,
-    )
+########################################################################
+# LeftAction python class TODO
```
Joseph-Edwards, line 324:
```suggestion
# LeftAction python class
```

libsemigroups_pybind11/stephen.py
```diff
@@ -9,68 +9,120 @@
...
     StephenPresentationWords as _StephenPresentationWords,
     StephenInversePresentationWords as _StephenInversePresentationWords,
-    accepts,
-    dot,
-    is_left_factor,
-    left_factors,
-    number_of_left_factors,
-    number_of_words_accepted,
-    words_accepted,
+    # TODO rename with stephen_ prefix
```
Joseph-Edwards, line 24: Is this todo necessary since the names are unique within this submodule?

libsemigroups_pybind11/to.py
```diff
@@ -10,8 +10,8 @@
 
 from typing import List, _GenericAlias
 from _libsemigroups_pybind11 import (
-    to_congruence_string,
-    to_congruence_word,
+    to_congruence_string as _to_congruence_string,
+    to_congruence_word as _to_congruence_word,
     to_froidure_pin,
     to_inverse_presentation_string,
     to_inverse_presentation_word,
```
Joseph-Edwards, line 17: Should these be underscored as well?

src/present.cpp
```diff
@@ -32,7 +32,7 @@
 
 // libsemigroups....
 #include <libsemigroups/constants.hpp>     // for operator==, UNDEFINED
-#include <libsemigroups/exception.hpp>     // for LibsemigroupsException
+#include <libsemigroups/exception.hpp>     // for LibsemigroupsError
```
Joseph-Edwards, line 35:
```suggestion
#include <libsemigroups/exception.hpp>     // for LibsemigroupsException
```

Joseph-Edwards (changes requested):
This PR looks awesome; thanks @james-d-mitchell! I've made a few minor suggestions below. Some will be quick fixes, but others may require a bit of discussion about the best course of action. Overall looks great :)

libsemigroups_pybind11/to.py
james-d-mitchell: Yeah

libsemigroups_pybind11/action.py
james-d-mitchell: I think I've fixed all of these now, should be good to go! Thanks @Joseph-Edwards

Joseph-Edwards:
I think it would be very helpful to have a way of formatting the rst code, but some of the changes introduced in 2fc2c51 don't interact with Sphinx's autodoc stuff very well.

For example, in [paths.rst](https://github.com/libsemigroups/libsemigroups_pybind11/pull/235/files#diff-516aa1c9c44ed81e59046a8ca12032ffd0a748e45719202d885a2939c8a62d55R20-R33), the entries in the `autosummary` are indented by 10 spaces, some files use five spaces, and most use 4.

It looks like most of the stuff has been formatted well though.

libsemigroups_pybind11/adapters.py
```diff
@@ -6,114 +6,152 @@
...
+from typing import TypeVar as _TypeVar
+from typing_extensions import Self as _Self
 
-from _libsemigroups_pybind11 import (
-    ImageRightActionBMat8BMat8 as _ImageRightActionBMat8BMat8,
-    ImageLeftActionBMat8BMat8 as _ImageLeftActionBMat8BMat8,
-    ImageRightActionPPerm1PPerm1 as _ImageRightActionPPerm1PPerm1,
-    ImageLeftActionPPerm1PPerm1 as _ImageLeftActionPPerm1PPerm1,
-    ImageRightActionPPerm1List as _ImageRightActionPPerm1List,
+from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module,unused-import
```
Joseph-Edwards, line 17:
```suggestion
from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module
```
since we do use all of the imports.

libsemigroups_pybind11/adapters.py
```diff
@@ -6,114 +6,152 @@
...
+    to_py as _to_py,
+    copy_cxx_mem_fns as _copy_cxx_mem_fns,
+    register_cxx_wrapped_type as _register_cxx_wrapped_type,
+)
 
-from .tools import ordinal
-from .transf import PPerm
+from .detail.decorators import copydoc as _copydoc
 
+from .transf import PPerm as _PPerm
```
Joseph-Edwards, line 39:
```suggestion
```

libsemigroups_pybind11/congruence.py
```diff
@@ -6,119 +6,132 @@
...
+manipulating :any:`Congruence` objects. All such functions are contained in
 the submodule ``congruence``.
 """
 
-from typing import List, Iterator, Union
+from typing import List, Union
+from typing_extensions import Self as _Self
 
-from _libsemigroups_pybind11 import (
+from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module,unused-import
```
Joseph-Edwards, line 18:
```suggestion
from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module
```

libsemigroups_pybind11/todd_coxeter.py
```diff
@@ -6,14 +6,16 @@
...
+are contained in the subpackage ``todd_coxeter``.
+"""
 
-from typing import Union, List, Iterator
-from typing_extensions import Self
+from typing import List
+from typing_extensions import Self as _Self
 
-from _libsemigroups_pybind11 import (
+from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module,unused-import
```
Joseph-Edwards, line 18: We can get rid of `unused-imports` here if we also remove `PositiveInfinity`, `_PresentationStrings`, `_PresentationWords` and `Undefined`.

libsemigroups_pybind11/transf.py
```diff
@@ -6,63 +6,58 @@
...
 
-from typing import Any, Union
+from typing import Any as _Any, List, Union
 from typing_extensions import Self
 
-from _libsemigroups_pybind11 import (
-    Transf1 as _Transf1,
-    Transf2 as _Transf2,
-    Transf4 as _Transf4,
+from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module,unused-import
```
Joseph-Edwards, line 19:
```suggestion
from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module
```

Joseph-Edwards (commented):
Thank you for making the changes @james-d-mitchell. I think I've spotted a few more places where we can remove `unused-imports`. Otherwise, once the tests pass, I'm happy to merge this.

james-d-mitchell:
> I think it would be very helpful to have a way of formatting the rst code, but some of the changes introduced in [2fc2c51](https://github.com/libsemigroups/libsemigroups_pybind11/commit/2fc2c512b6cf00f8248294dfc11c510ea78012af) don't interact with Sphinx's autodoc stuff very well.
> 
> For example, in [paths.rst](https://github.com/libsemigroups/libsemigroups_pybind11/pull/235/files#diff-516aa1c9c44ed81e59046a8ca12032ffd0a748e45719202d885a2939c8a62d55R20-R33), the entries in the `autosummary` are indented by 10 spaces, some files use five spaces, and most use 4.
> 
> It looks like most of the stuff has been formatted well though.

Hmmm I’m not seeing any warnings when I run make doc, are you? If so, can you please post them? Thanks!

Joseph-Edwards:
> Hmmm I’m not seeing any warnings when I run make doc, are you? If so, can you please post them? Thanks!

I don't get any warnings either. It just seemed odd that the `.rst` files now seem to have inconsistent amounts of indentation.

As far as I can tell, this hasn't changed the appearance of the doc itself.

james-d-mitchell:
> > Hmmm I’m not seeing any warnings when I run make doc, are you? If so, can you please post them? Thanks!
> 
> I don't get any warnings either. It just seemed odd that the `.rst` files now seem to have inconsistent amounts of indentation.
> 
> As far as I can tell, this hasn't changed the appearance of the doc itself.

Ah right, I think it was pretty inconsistent before too, so I'm not sure it's now worse than before. I should file an issue with `docstrfmt`, also there were some changes that it made that did break the doc, which I fixed.

Are you happy for this to be merged?

Joseph-Edwards:
> Ah right, I think it was pretty inconsistent before too, so I'm not sure it's now worse than before. I should file an issue with `docstrfmt`, also there were some changes that it made that did break the doc, which I fixed.
> 
> 
> 
> Are you happy for this to be merged? 

Yeah in that case I'm happy to merge. Thanks again @james-d-mitchell

PR #236, opened 2025-04-02
Update documentation building process

Joseph-Edwards:
This PR moves all of our ~doc hacks~ custom sphinx behaviour into an extension, thus reserving `conf.py` for configuration information only. It also checks which string replacements we actually use.

We only check which string replacements we use when the documentation is built from a fresh environment (i.e. not an incremental build) to avoid false negatives. As such, this PR also adds a make recipe `fresh-doc` which behaves like `make clean-doc && make doc`, whilst also passing an extra flag to `etc/make-doc.sh` that enables the string-replacement-checking for the post-processing of the html files.


james-d-mitchell:
Looks good thanks @Joseph-Edwards

PR #237, opened 2025-04-03
Modernise the build system

Joseph-Edwards:
This PR modernises the build system, and makes the package adhere some Python good practices. Specifically, this PR does the following:

- [X] Move as much project metadata out of `setup.py` into `pyproject.toml` as possible (see the [Python Packaging Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) and [PEP 518](https://peps.python.org/pep-0518/)).
- [X] Move some of the tools used for building `libsemigroups_pybind11` into their own module to prevent circular includes.
- [X] Refactor the code code that finds include directories and library directories.
- [X] Use `setuptools-scm` to use git information to define the version of the module (see [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/)).
- [X] Consolidate and update the various `environment.yml` and `requirements.txt` files.
- [X] Move `libsemigroups_pybind11/` into `src/` (see [src layout vs flat layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/)).
- [ ] Update the list of authors and contributors.


Joseph-Edwards:
@james-d-mitchell, this feels like a good point to specify authors and maintainers in the `pyproject.toml`. Presently, I've just moved them over from the old `setup.py` list, but we should probably update that. Are you happy for me to leave this decision to you?

james-d-mitchell:
I have some questions, which we can discuss next week, but this looks really good

Joseph-Edwards:
This has now been rebased on top of the latest v1 changes. Other than updating the list of authors, which could be deferred to a different PR, I think this is good to go.

Makefile
```diff
@@ -39,8 +37,8 @@ clean-doc:
 	rm -rf docs/_build
 
 clean: clean-doc
-	rm -rf __pycache__ libsemigroups_pybind11.egg-info
-	rm -rf tests/__pycache__ libsemigroups_pybind11/__pycache__
+	find . -type d -name __pycache__ -prune -exec rm -rf {} \;
```
james-d-mitchell, line 40: much better

etc/make-lint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
james-d-mitchell, line 1: I suppose this intentionally doesn't have `set -e` at the start so that it continues even if something fails?

etc/make-lint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [[ $# -ne 0 ]]; then
+    bold "error expected 0 arguments, got $#!"
+    exit 1
+fi
+
+exit_code=0
+
+echo "Linting with ruff . . ."
+ruff check setup.py tests/*.py src/libsemigroups_pybind11/*.py docs/source/*.py || ((exit_code = 1))
```
james-d-mitchell, line 11: Not sure why ruff and pylint are run on different files? In particular, looks like we're not running ruff on anything in `src/libsemigrous_pybind11/detail` and not running pylint on `docs/source/*.py`??

james-d-mitchell (approved):
Looks good and happy to merge apart from the one (possible?) issue in the make-lint script.

etc/make-lint.sh
Joseph-Edwards: Yeah that's right. Then we test for non-zero exit status as we go, but only exit out once all of the linting steps are done.

src/libsemigroups_pybind11/detail/cxx_wrapper.py
Joseph-Edwards, line 1: According to ruff:
```
src/libsemigroups_pybind11/detail/cxx_wrapper.py:61:7: B024 `CxxWrapper` is an abstract base class, but it has no abstract methods or properties
   |
61 | class CxxWrapper(metaclass=abc.ABCMeta):
   |       ^^^^^^^^^^ B024
62 |     # pylint: disable=missing-class-docstring
63 |     # pylint: disable=protected-access, no-member
   |
```

@james-d-mitchell, do you think we should suppress this, or do something about it?

Joseph-Edwards: As a side not, running `ruff rule B024` gives a really nice explanation of the rule. Seems like a cool feature of ruff.

james-d-mitchell: Just remove the abstract class please, it’s just a left over

PR #241, opened 2025-04-30
added #include <libsemigroups/word-graph-helpers.hpp>

nadim-241:
PyBind checks in libsemigroups were not compiling due to missing #include <libsemigroups/word-graph-helpers.hpp>. This adds this include so that the library builds.


james-d-mitchell:
Super thanks @nadim-241

PR #258, opened 2025-06-05
Use the correct error types

Joseph-Edwards:
This PR ensures the correct errors are being thrown, and that we document the errors that actually get thrown. It also updates how we register our custom `LibsemigroupsError` to move away from some deprecated functions.

There are a few places where we arguably throw the wrong error (for example, we throw a `LibsemigroupsError` rather than a `ValueError` when trying to add a generator to a `FroidurePin` with the wrong degree). We could fix this, but this seems like more hassle than it's worth.

### DIFF
--- a/src/errors.cpp
+++ b/src/errors.cpp
@@ -57,19 +57,23 @@ namespace libsemigroups {
   void init_error(py::module& m) {
     // TODO this doesn't seem to properly catch all LibsemigroupsExceptions,
     // particularly on macOS. This may have been resolved in pybind11 2.12.0
-    static py::exception<LibsemigroupsException> exc(
-        m, "LibsemigroupsError", PyExc_RuntimeError);
+
+    // Using the GIL safe call below rather than simply having a static
+    // py::exception is recommended in the pybind11 doc.
+    PYBIND11_CONSTINIT static py::gil_safe_call_once_and_store<py::object>
+        exc_storage;
+    exc_storage.call_once_and_store_result([&]() {
+      return py::exception<LibsemigroupsException>(
+          m, "LibsemigroupsError", PyExc_RuntimeError);
+    });
     py::register_exception_translator([](std::exception_ptr p) {
       try {
         if (p) {
           std::rethrow_exception(p);
         }
       } catch (LibsemigroupsException const& e) {
-        exc(formatted_error_message(e).c_str());
-      } catch (py::stop_iteration const& e) {
-        throw e;
-      } catch (std::runtime_error const& e) {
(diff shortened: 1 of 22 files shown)


src/errors.cpp
```diff
@@ -57,25 +57,56 @@ namespace libsemigroups {
...
+
     m.def("error_message_with_prefix",
-          py::overload_cast<>(&error_message_with_prefix));
+          py::overload_cast<>(&error_message_with_prefix),
+          R"pbdoc(
+Return whether :any:`LibsemigroupsError` messages have a C++ prefix.
+
+Since ``libsemigroups_pybind11`` is built on top of the C++ library
+``libsemigroups``, many of the errors thrown in ``libsemigroups_pybind11``
+emanate from a C++ function. This function returns whether
```
james-d-mitchell, line 87:
```suggestion
emanate from C++. This function returns whether
```

tests/test_words.py
```diff
@@ -348,5 +348,5 @@ def test_human_readable_letter():
 
 def test_human_readable_index():
     assert human_readable_index("\377") == 255
-    with pytest.raises(LibsemigroupsError):
```
james-d-mitchell, line 351: I'm a bit confused about what raises the ValueError here? isn't `human_readable_index` a C++ function in libsemigroups? If so, how is it raising a `ValueError`?

Joseph-Edwards: It's because the C++ function is defined on chars that take values from 0 to 255, but python has no char type. So the parameter type for this function in python is `str`, but it seems as though some work is done to make sure the string you pass can be represented as a C++ `char`, and throws a value error if it can't. From what I can tell, it checks if the string has length one, and that it is in the range [0, 256). This happens before our C++ gets called. E.g.:
```py
In [20]: human_readable_index(chr(256))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[20], line 1
----> 1 human_readable_index(chr(256))

ValueError: Character code point not in range(0x100)
```
or
```py
In [24]: human_readable_index("ab")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[24], line 1
----> 1 human_readable_index("ab")

ValueError: Expected a character, but multi-character string found
```

Joseph-Edwards:
Once https://github.com/libsemigroups/libsemigroups/pull/736 is merged, this PR should be good to go from my perspective.

Joseph-Edwards:
Upon further consideration, this PR seems like a good place to include the "describe the `_libsemigroups_pybind11` module" changes, so I'll do that now too.

PR #261, opened 2025-06-09
Homogenize the doc

Joseph-Edwards:
This PR addresses some inconsistencies within in the doc. Specifically:

- [X] The "Elements" section now has the same structure as the rest of the doc;
- [X] Class pages all follow the same structure.
- [X] All functions now appear in autosummaries


Joseph-Edwards:
This PR will likely be the first of several doc-related PRs, rather than having a big behemoth.

james-d-mitchell:
@Joseph-Edwards it's probably better if we merge smaller chunks rather than bigger ones, so I'd be happy enough to merge this, and have the remaining check boxes as separate PRs, if that's ok with you?

PR #265, opened 2025-06-11
runner: add dummy base classes Runner + Reporter

james-d-mitchell:


james-d-mitchell:
We decided not to go down this route, so closing.

PR #266, opened 2025-06-12
doc: readd MatrixKind doc

james-d-mitchell:
Seems to have been accidentally removed.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -5,7 +5,7 @@
 
     The full license is in the file LICENSE, distributed with this software.
 
-.. |libsemigroups-pybind11-version| replace:: 1.0.0
+|libsemigroups-pybind11-version| replace:: 1.0.0
 
 Installation
 ============
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -5,7 +5,7 @@
 
     The full license is in the file LICENSE, distributed with this software.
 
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
@Joseph-Edwards suggests moving the doc to the python file, and formatting it like the `tril` enum would be good, and I agree, so I should do that.

james-d-mitchell:
> @Joseph-Edwards suggests moving the doc to the python file, and formatting it like the `tril` enum would be good, and I agree, so I should do that.

Now done.

docs/source/install.rst
```diff
@@ -5,7 +5,7 @@
 
     The full license is in the file LICENSE, distributed with this software.
 
-.. |libsemigroups-pybind11-version| replace:: 1.0.0
+|libsemigroups-pybind11-version| replace:: 1.0.0
```
Joseph-Edwards, line 8: I think this causes
```
/home/runner/work/libsemigroups_pybind11/libsemigroups_pybind11/docs/source/install.rst:8: ERROR: Undefined substitution referenced: "libsemigroups-pybind11-version". [docutils]
```

Joseph-Edwards, line 8:
```suggestion
.. |libsemigroups-pybind11-version| replace:: 1.0.0
```

Joseph-Edwards (changes requested):

PR #267, opened 2025-06-13
Doc cleanups

james-d-mitchell:


src/knuth-bendix.cpp
```diff
@@ -332,15 +332,21 @@ redundant in this way, then ``None`` is returned.
...
+            if (it != p.rules.cend()) {
+              return std::make_pair(*it, *(it + 1));
+            }
+            return {};
+          },
+          // Signature required to avoid generating any doc
+          R"pbdoc(
+:sig=(p: Presentation, t: datetime.timedelta) -> tuple[list[int], list[int]] | tuple[str, str] | None:
+:only-document-once:
+)pbdoc");
```
Joseph-Edwards, line 349: This fixes the doc, but this function probably shouldn't be bound in `bind_knuth_bendix`, since `redundant_rule` doesn't depend on the template arguments. Hence, we end up binding the functions 16 times, rather than twice. I will fix this in a separate PR.

src/todd-coxeter-impl.cpp
```diff
@@ -119,8 +119,9 @@ The values in this enum can be used as the argument for
 should be performed.)pbdoc")
         .value(
             "hlt",
+            // JDM couldn't get the links to work in the following doc string.
```
Joseph-Edwards, line 122: I think if the enums were written like tril, we could get this to work. I'll give it a go, and open a new PR if it works. Otherwise, I think this is fine.

Joseph-Edwards (approved):
Thanks @james-d-mitchell. There are a few things that this PR highlighted, but I will fix those in separate PRs.

PR #270, opened 2025-06-13
CI: Add a job that checks the doc builds without warnings

Joseph-Edwards:
This PR adds a job that checks the doc builds without warnings. It also factors out some duplicated code in the run tests job.


Joseph-Edwards:
This job could very easily become a step in the "Run tests and lint" job, which would mean we don't have to make libsemigroups again, but I think having separate jobs is good.

PR #275, opened 2025-06-16
Reorganise TOC

Joseph-Edwards:
This PR moves some files around so that:
 -  there is a "Package Info" with authors, biblio, installation and changelog;
 -  there is a "Core classes" section for `Runner` and `Reporter` and `ReportGuard`;
 -  there is a section for constants, that has been corrected to remove references to hidden types;
 - there is an enum page that documents all module level enums.

The enum doc still need to be normalised, but this will be done in a later PR.


james-d-mitchell:
Sweet thanks @Joseph-Edwards

PR #277, opened 2025-06-17
Doc nested classes at same depth as parent

james-d-mitchell:


james-d-mitchell:
> This all looks good to me; however, I have checked out this branch to look at the doc and for some reason I can't pip install after rebasing on top of v1. I'll see if I can determine why. If it seems like a me issue, I'm happy to merge

Probably you need to `make install` libsemigroups after pulling from `main`.

PR #281, opened 2025-06-17
Fix copy of non-function attributes

james-d-mitchell:


Joseph-Edwards (approved):
Looks good. Thanks @james-d-mitchell

PR #285, opened 2025-06-19
Update install rst

james-d-mitchell:


docs/source/install.rst
```diff
@@ -13,51 +13,53 @@ Installation
...
 
-It's possible to install ``libsemigroups_pybind11`` using ``pip`` via one of:
+It's possible to install ``libsemigroups_pybind11`` using ``pip`` something
+like the following (depending on your system and setup):
 
 ::
 
     pip install libsemigroups_pybind11
     pip3 install libsemigroups_pybind11
     python -m pip install libsemigroups_pybind11
```
Joseph-Edwards, line 23: I think this reads as thought you need to run all three lines. Perhaps three separate code blocks would be better?

docs/source/install.rst
```diff
@@ -13,51 +13,53 @@ Installation
...
+        At present this does not work for Macs with M1 processors.
 
 From the sources
 ----------------
 
 Before installing ``libsemigroups_pybind11`` from its sources you should first
 perform a system install of the C++ library ``libsemigroups``. For information
 about how to install ``libsemigroups`` see `the installation guide
-<https://libsemigroups.readthedocs.io/en/latest/install.html>`_.
+<https://libsemigroups.github.io/libsemigroups/md_install.html>`_.
```
Joseph-Edwards, line 62: This might be clearer as: "For information about how to do this, see [the libsemigroups installation guide](https://libsemigroups.github.io/libsemigroups/md_install.html)"

docs/source/install.rst
Joseph-Edwards, line 1: In the "From a release archive" section, the links are wrong because of some leftover string replacement stuff. (I can't comment on those lines directly because they weren't changed in this PR)

docs/source/install.rst
```diff
@@ -84,21 +86,12 @@ To build ``libsemigroups_pybind11`` from a release archive:
...
+Assuming you already have ``python3`` install, you can build the doc by using:
 
 ::
 
-    python3 -m pip3 install -r docs/requirements
-
-Then it ought to be possible to just run ``make doc`` in the ``libsemigroups``
-directory.
+    python3 -m pip3 install -r requirements.txt
+    make doc
```
Joseph-Edwards, line 94: Do we expect all of our target machines to have some form of `automake` / a tool that can execute `make doc`? If yes, this is fine. If no, this could be replaced with
``` bash
./etc/make-doc.sh
```

Joseph-Edwards (changes requested):
Looks mostly good to me. I have made a few stylistic comments, and pointed out some small mistakes, but otherwise good to me. Thanks @james-d-mitchell

docs/source/install.rst
Joseph-Edwards: Yeah that seems sensible.

PR #286, opened 2025-06-19
doc: more properly say that Runners don't subclass Runner

james-d-mitchell:


src/runner.cpp
```diff
@@ -56,13 +56,26 @@ The time between the given point and now.
...
 -  :any:`report_prefix()`;
 -  :any:`report_every()`;
 -  :any:`last_report()`;
 -  :any:`start_time()`.
+
+This class has limited (no?) utility on its own, and you are unlikely to want to use it directly.
+
+.. important::
+
+   The classes derived from :any:`Reporter` have all the methods of the
```
Joseph-Edwards, line 71:
```suggestion
   The classes that we claim derive from :any:`Reporter` in this documentation have all the methods of the
```

Joseph-Edwards: And something similar for runner too?

Joseph-Edwards (commented):
Everything looks good to me. I think we could potentially reword the notice in `Runner` and `Reporter` to make a distinction between "derived from" and "what we claim derives from". But if not, this is good to go.

src/runner.cpp
```diff
@@ -68,7 +68,7 @@ This class has limited (no?) utility on its own, and you are unlikely to want to
 
 .. important::
 
-   The classes derived from :any:`Reporter` have all the methods of the
+   The classes that we claim derive from :any:`Reporter` in this documentation have all the methods of the
```
Joseph-Edwards, line 71: Should we make a similar change for `Runner` too?

PR #295, opened 2025-07-03
knuth-bendix: expose `internal_presentation`

james-d-mitchell:
`internal_presentation` is available for `ToddCoxeter` so I'm adding it to `KnuthBendix` too. This is helpful mostly for debugging.


james-d-mitchell:
Requires #298 to be merged first

PR #297, opened 2025-07-04
Doc consistency

Joseph-Edwards:
This PR adds checks that the docstring is in a consistent order, and that the parameters in the signature have the same name and typehints as the parameters in the docstrings. It also fixes any warnings that arise as a result.


docs/source/data-structures/presentations/to-present.rst
```diff
@@ -207,9 +207,8 @@ enumerates *kb*) prior to calling this function.
     >>> kb.run()
     >>> p2 = to(kb, rtype=(Presentation,))
     >>> for p in [p1, p2]:
-    ...     presentation.sort_each_rule(p)
-    ...     presentation.sort_rules(p)
-
+    ...     _ = presentation.sort_each_rule(p)
```
james-d-mitchell, line 210: Pretty sure that this function and the one in the next line return `None` what's the point of this change?

Joseph-Edwards: The `libsemigroups` functions `sort_each_rule(Presentation<Word>& p)` and `sort_each_rule(Presentation<Word>& p, Compare cmp)` both return whether any rules were sorted. Prior to this PR, the python analogue of `sort_each_rule(Presentation<Word>& p, Compare cmp)` returned a bool, but `sort_each_rule(Presentation<Word>& p)`  discarded its return type for some reason. I changed it so they both return a bool now.

That said, `sort_rules` does indeed return `None` (which is perhaps that it isn't the same as `sort_each_rule`), so `_ = ` for that one probably shouldn't be there.

Joseph-Edwards: Since `sort_each_rule` now returns a bool, the doctest now expects some output. Adding `_ =` was a means of suppressing this.

james-d-mitchell: Aha, so why not include the output?

Joseph-Edwards: Yeah that's probably the best thing to do

PR #311, opened 2025-08-11
Fix some broken links

Joseph-Edwards:
This PR fixes some broken links in the doc.


Joseph-Edwards:
In the [Numpy 2.0.2 release notes](https://numpy.org/devdocs/release/2.0.2-notes.html), it's claimed that the supported versions of Python are 3.9-3.12. I'll have a mess around with the different versions and see if this fixes things.

We already conditionally install different versions of dependencies based on the version of Python being used, and I see no issue of doing that here too if necessary.

james-d-mitchell (approved):

PR #240, opened 2025-04-30
Use conda version of `libsemigroups` in the CI

Joseph-Edwards:
This PR updates the CI to use the conda version of `libsemigroups`, and also adds a job that uses the main version of `libsemigroups`.

PR #322, opened 2025-09-02
Add CI workflow that tests using conda `libsemigroups`

Joseph-Edwards:
This PR adds a workflow that tests `libsemigroups_pybind11` against the conda version of `libsemigroups`. It also moves the linting into its own workflow.

Supersedes #240.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -32,7 +32,6 @@ jobs:
           environment-name: libsemigroups
           create-args: >-
             python
-            fmt
       - name: "Set environment variables . . ."
         run: |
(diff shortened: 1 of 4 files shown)


Joseph-Edwards:
The Conda tests are failing because we haven't made a release of libsemigroups since adding the `balance` (and friends) functions. This is expected.

Arguably, it is too aggressive to run the tests for Conda and GitHub whenever we make a PR, since we might not mind if some of them fail. We could instead only check against GitHub libsemigroups for PRs into main, and against Conda libsemigroups for PRs into Stable branches.

On the other hand, it serves as a good reminder to make regular small releases, so I'm not sure what's best.

Joseph-Edwards:
The reason for testing all of the versions is for making sure the Python code we've written is compatible with all of the versions of Python. For example, when we started adding typing support, we needed to check it was valid for the older versions of Python.

That said, I agree that we don't need to cheek all of the versions twice.

I think it's probably best to check all of the versions against GitHub libsemigroups, and only the conda one once. What do you think @james-d-mitchell?

PR #324, opened 2025-09-05
Test default values and `datetime` types

Joseph-Edwards:
This PR adds some tests of the default values of `ToddCoxeter` settings, and also the type of `datetime` object some functions return.

### DIFF
--- a/src/todd-coxeter-impl.cpp
+++ b/src/todd-coxeter-impl.cpp
@@ -969,6 +969,8 @@ lookaheads to be stopped early if the number of nodes being killed is too small
 want to stop the lookahead early, since lookaheads take some time but may not
 result in many nodes being killed).
 
+The default value is `0.01`
+
 :param val: the proportion of active nodes.
 :type val: float
 
@@ -1018,7 +1020,9 @@ The default value of this setting is ``options.lookahead_style.hlt``.
 )pbdoc");
     thing.def(
         "lower_bound",
-        [](ToddCoxeterImpl_ const& self) { return self.lower_bound(); },
+        [](ToddCoxeterImpl_ const& self) {
+          return from_int(self.lower_bound());
+        },
         R"pbdoc(
 :sig=(self: ToddCoxeter) -> int:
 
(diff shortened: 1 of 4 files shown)


src/runner.cpp
```diff
@@ -31,11 +35,57 @@
...
+
+    template <
+        typename TimePoint,
+        std::enable_if_t<!std::is_same_v<TimePoint, high_resolution_clock>,
+                         bool>
+        = true>
+    system_clock::time_point to_system(TimePoint tp) {
+      // Account for the difference between system_clock and
+      // high_resolution_clock
+      auto sys_now      = system_clock::now();
```
james-d-mitchell, line 58: I'm not sure this does what you want, the time at which line 58 is executed isn't the same as the time at which 59 is executed, so what are you measuring the here?

src/runner.cpp
```diff
@@ -31,11 +32,48 @@
...
+        return time_point_cast<system_clock::duration>(tp);
+      } else {
+        // Account for the difference between high_resolution_clock and
+        // system_clock
+        auto sys_now   = system_clock::now();
+        auto clock_now = high_resolution_clock::now();
+        return time_point_cast<system_clock::duration>(tp - clock_now
+                                                       + sys_now);
+      }
+    }
```
Joseph-Edwards, line 50: This is what I would like to do, but this causes a compiler error on Mac because:
```
no viable conversion from returned value of type 'time_point<std::chrono::steady_clock, [...]>' to function return type 'time_point<system_clock, [...]>'
```
on line 41, even though the `if constexpr` should prevent this line from being compiled (or so I thought).

Joseph-Edwards: I understand why this isn't compiling now, but still struggling to get the behaviour I desire. I will persevere

Joseph-Edwards: The reason it wasn't compiling was because my constexpr if was not inside a template. Therefore, each branch was fully checked by the compiler, which resulted in errors. I think I've fixed this now in 08d4a65.

PR #343, opened 2025-11-04
Use uv in favour of pip

Joseph-Edwards:
Before this PR, we used `requirements.txt` to try and create reproducible environments for our CI to run in. These files are non-standardized, awkward to maintain, and give different behaviour on different systems.

This PR adds uv to our workflow to address these issues. The `uv.lock` file does the work of the old `requirements.txt` file, and is automatically generated (by running `uv lock`) using the information we already specify in `pyproject.toml`. We now use uv instead of pip in our CI.

I haven't purged `requirements.txt` and its associated scripts from the project yet, in case there are any users that use them, but I'm happy to do so if we don't think that's an issue

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -26,6 +26,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
 
       # Used to host cibuildwheel
       - name: Build wheel


Joseph-Edwards:
When you say a job using using pip, do you mean:
1. a job that uses `pip install .` to install `libsemigroups_pybind11`;
2. a job that installs the dependencies through `pin install -r requirements.txt`; or
3. both?

I think case 2 (and therefore 3) would be a bit overkill, because that would require maintaining the list of dependencies which this PR aims to remove.

1 would be doable quite easily, but under the hood both pip and uv will be invoking setuptools to build and install the project (I think). As long as both of these tools remain [PEP 517](https://peps.python.org/pep-0517/) compliant, I don't think their behaviour in this particular instance can deviate too far. That said, that's not a reason to not add a pip job.

What do you think @james-d-mitchell?

Joseph-Edwards:
Going to temporarily close this whilst I debug the CI

PR #347, opened 2025-11-07
Sync `Forest` and its helpers with recent changes in `libsemigroups`

james-d-mitchell:
Will require

https://github.com/libsemigroups/libsemigroups/pull/847

to be merged before the ci here will pass.

### DIFF
--- a/src/libsemigroups_pybind11/forest.py
+++ b/src/libsemigroups_pybind11/forest.py
@@ -14,7 +14,7 @@ the submodule ``forest``.
 
 from typing_extensions import Self as _Self
 
-from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module
+from _libsemigroups_pybind11 import (  # pylint: disable=no-name-in-module, unused-import
     forest_PathsFromRoots as PathsFromRoots,
     forest_PathsToRoots as PathsToRoots,
     forest_depth as depth,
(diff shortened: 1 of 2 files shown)


src/forest.cpp
```diff
@@ -341,5 +342,533 @@ Set the parent and edge label for a node. This function sets the parent of
...
+:returns: A copy.
+:rtype: PathsToRoots
+)pbdoc");
+
+      thing.def(
+          "__iter__",
+          [](forest::PathsToRoots& self) {
+            return py::make_iterator(self.begin(), self.end());
+          },
+          py::keep_alive<0, 1>(),
```
Joseph-Edwards: I would benefit from a comment that explains what this slightly cryptic thing is doing, but maybe I'm just being a noob. My understanding is that this means "Keep `this` alive for at least as long as the returned iterator is alive"?

Feel free to disregard if you think this is clear as it is.

Joseph-Edwards (changes requested):
This looks good to me. I've requested one small doc change, and made one comment, but otherwise I'm happy for this to be merged.

It's worth noting that we expect this to fail the conda libsemigroups jobs. This raises the question that maybe they should only be run when making PRs into `stable-*`, or on a release? What do you think @james-d-mitchell?

james-d-mitchell:
> This looks good to me. I've requested one small doc change, and made one comment, but otherwise I'm happy for this to be merged.
> 
> It's worth noting that we expect this to fail the conda libsemigroups jobs. This raises the question that maybe they should only be run when making PRs into `stable-*`, or on a release? What do you think @james-d-mitchell?

I think that'd be a good idea, this way we'll only see proper errors, not out of sync issues like this.

PR #350, opened 2025-11-07
paths: add helpers

james-d-mitchell:


src/paths.cpp
```diff
@@ -404,5 +404,352 @@ The underlying word graph. This function returns underlying
...
+  is the out-degree of the word graph.
+
+:note:
+ If ``libsemigroups`` is compiled with the flag ``--enable-eigen``, then
+ this function makes use of the `Eigen` library for linear algebra (see
+ :cite:`Guennebaud2010aa`).
+
+:warning:
+  If the number of paths exceeds ``2 ** 64``, then return value of
+  this function will not be correct.
```
Joseph-Edwards:
```suggestion
.. warning::
    If the number of paths exceeds ``2 ** 64``, then return value of
    this function will not be correct.
```

src/paths.cpp
```diff
@@ -404,5 +404,352 @@ The underlying word graph. This function returns underlying
...
+  * :any:`paths.algorithm.trivial`: at worst :math:`O(nm)` where :math:`n` is the number
+    of nodes and :math:`m` is the out-degree of the word graph (only valid in
+    some circumstances)
+  * :any:`paths.algorithm.automatic`: attempts to select the fastest algorithm of the
+    preceding algorithms and then applies that.
+
+:note:
+ If ``libsemigroups`` is compiled with the flag ``--enable-eigen``, then
+ this function makes use of the `Eigen` library for linear algebra (see
+ :cite:`Guennebaud2010aa`).
```
Joseph-Edwards:
```suggestion
.. note::
    If ``libsemigroups`` is compiled with the flag ``--enable-eigen``, then
    this function makes use of the `Eigen` library for linear algebra (see
    :cite:`Guennebaud2010aa`).
```

src/paths.cpp
```diff
@@ -404,5 +404,352 @@ The underlying word graph. This function returns underlying
...
+
+    ////////////////////////////////////////////////////////////////////////
+    // Helper functions
+    ////////////////////////////////////////////////////////////////////////
+
+    m.def(
+        "paths_number_of_paths_algorithm",
+        [](WordGraph<uint32_t> const& wg, uint32_t source) {
+          return number_of_paths_algorithm(wg, source);
+        },
```
Joseph-Edwards: For consistency with the rest of this file, I think it would be better (albeit equivalent) to use `node_type` instead of `uint32_t` here and in the following functions.

Joseph-Edwards (changes requested):
Some minor doc and consistency comments, but otherwise this looks good to me. Thanks @james-d-mitchell :)

PR #351, opened 2025-11-10
Sync presentation

james-d-mitchell:
This PR adds support for 3 helper functions from `libsemigroups` that were not supported before. I think this means that all helper functions that we want to support from `presentation` are now supported.

### DIFF
--- a/src/present.cpp
+++ b/src/present.cpp
@@ -1483,6 +1483,40 @@ and *rhs* is the next item in ``p.rules``.
   if :any:`Presentation.throw_if_bad_alphabet_or_rules` throws.
 )pbdoc");
 
+      m.def(
+          "presentation_try_detect_inverses",
+          [](Presentation<Word>& p) {
+            return presentation::try_detect_inverses(p);
+          },
+          py::arg("p"),
+          R"pbdoc(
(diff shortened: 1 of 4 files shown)


src/present.cpp
```diff
@@ -1431,6 +1431,93 @@ This function returns the minimum index ``i`` of *lhs* such that ``p.rules[i +
...
+  A tuple where the first item consists of letters such that an inverse was
+  detected; and the second item consists of the inverses such that the item in
+  position ``i`` is the inverse of the letter in position ``i``.
+:rtype: tuple[str | list[int], str | list[int]]
+
+:raises LibsemigroupsError:
+  if :any:`throw_if_bad_alphabet_or_rules` throws.
+:raises LibsemigroupsError:
+  if conflicting inverses for any letter are detected.)pbdoc");
+
```
Joseph-Edwards:
```suggestion
```

Joseph-Edwards (changes requested):
This all looks good to me, mod the removal of the empty line that cpplint is complaining about.

A few of our recent PRs have been introducing whitespace/formatting changes on otherwise unchanged lines. This is fine for this PR, but we should probably discuss and settle on a common formatter to avoid this moving forward.

PR #353, opened 2025-11-11
Sync examples

james-d-mitchell:


docs/source/data-structures/presentations/examples.rst
```diff
@@ -75,6 +77,7 @@ documented in the :ref:`default-presentations` section.
 .. autosummary::
     :signatures: short
 
+    abacus_jones_monoid_AJP25
```
Joseph-Edwards: There should probably also be a `braid_group_Bir05`, but that's an issue for `libsemigruops`

PR #395, opened 2026-02-24
presentation: attempt to resolve #391

### LINKED ISSUE #391: Set item for `Presentation.rules` does nothing
If we do:

~~~
p = Presentation("abc")
p.rules = ["aaaa"]
p.rules[0] = "bbb"
p.rules
["aaaa"]
~~~

I think either we should make `p.rules` assignable (not sure if this is feasible) or it should throw an exception, silently failing like this caused me some confusion.

james-d-mitchell:
This resolves #391 by making `std::vector<word_type>` and `std::vector<std::string>` opaque types within `present.cpp` at least. The idea is that instead of such vectors being auto converted to lists, they are their own types. 

PROS: `p.rules` return a reference, (at least on my machine) the tests pass, and it's possible to do `p.rules[0] = something` and this actually works as expected. 

CONS: `p.rules` appears to be a `list`, but isn't, and so some things you might want to do with a `list` don't work (like `index` for example). This is a fairly large change.

I think the PROS out weigh the CONS, but happy to hear your opinion too @Joseph-Edwards

### DIFF
--- a/src/libsemigroups_pybind11/presentation/__init__.py
+++ b/src/libsemigroups_pybind11/presentation/__init__.py
@@ -11,8 +11,10 @@ from typing_extensions import Self as _Self
 from _libsemigroups_pybind11 import (
     InversePresentationString as _InversePresentationString,
     InversePresentationWord as _InversePresentationWord,
+    RulesWord as _RulesWord,
     PresentationString as _PresentationString,
     PresentationWord as _PresentationWord,
+    RulesString as _RulesString,
     presentation_add_cyclic_conjugates as _add_cyclic_conjugates,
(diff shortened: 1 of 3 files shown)


Joseph-Edwards:
I think that I initially objected to the changes in this PR because, as far as I can recall, none of our other classes have mutable data members so the proposed behaviour of the `rules` felt a bit unfamiliar. Upon reconsideration, this is not really a good reason to block these changes. As such, I am happy to go ahead with these changes (once they have been rebased and the CI passes).

james-d-mitchell:
Superseded by #505

PR #419, opened 2026-04-23
Consistent constants

Joseph-Edwards:
This PR includes some minor quality-of-life improvements such as removing redundant aliases, changing parameters to const-references, and using the `from_int` function more.

### DIFF
--- a/src/transf.cpp
+++ b/src/transf.cpp
@@ -111,7 +111,11 @@ the image of the point ``i`` under the {0} is ``imgs[i]``.
             // corresponds to a std::out_of_range for things like list(a) to
             // work.
             try {
-              return from_int(a.at(b));
+              auto result = a.at(b);
+              if (result != UNDEFINED) {
+                return {result};
+              }
+              return {UNDEFINED};
             } catch (LibsemigroupsException const& e) {
               throw std::out_of_range(formatted_error_message(e));
             }
@@ -167,7 +171,10 @@ definition, which is equal to the size of :any:`{0}.images`.
             auto r = rx::iterator_range(self.begin(), self.end())
                      | rx::transform(
                          [](auto val) -> int_or_unsigned_constant<Scalar> {
-                           return from_int(val);
+                           if (val != UNDEFINED) {
+                             return {val};
+                           }
+                           return {UNDEFINED};
                          });
             return py::make_iterator(rx::begin(r), rx::end(r));
           },
--- a/src/transf.cpp
+++ b/src/transf.cpp
@@ -111,11 +111,7 @@ the image of the point ``i`` under the {0} is ``imgs[i]``.
             // corresponds to a std::out_of_range for things like list(a) to
             // work.
             try {
-              auto result = a.at(b);
-              if (result != UNDEFINED) {
-                return {result};
-              }
-              return {UNDEFINED};
+              return from_int(a.at(b));
             } catch (LibsemigroupsException const& e) {
               throw std::out_of_range(formatted_error_message(e));
             }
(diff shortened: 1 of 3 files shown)


src/transf.cpp
```diff
@@ -171,10 +167,7 @@ definition, which is equal to the size of :any:`{0}.images`.
             auto r = rx::iterator_range(self.begin(), self.end())
                      | rx::transform(
                          [](auto val) -> int_or_unsigned_constant<Scalar> {
-                           if (val != UNDEFINED) {
-                             return {val};
-                           }
-                           return {UNDEFINED};
+                           return from_int(val);
                          });
```
Joseph-Edwards, line 183: It's possible to remove this lambda all together, with something like
```cpp
static_cast<int_or_constant<Scalar> (*)(Scalar)>(from_int)
```
but this felt a bit unwieldy to me, so I opted to keep the lambda.

Joseph-Edwards:
This is a draft until I can get the tests passing

james-d-mitchell:
> This is a draft until I can get the tests passing 

I don't see anything obviously wrong with the changes in this pr, I wonder what's going on with the ci

src/transf.cpp
james-d-mitchell: And you can't just replace the lambda with `from_int` ?

Joseph-Edwards: Unfortunately not; `transform` is unable to infer the type of `from_int`. I think this is because `from_int` is an overloaded function template. Something like this seems to work:
```cpp
auto r = rx::iterator_range(self.begin(), self.end())
         | rx::transform<int_or_constant<Scalar> (*)(Scalar)>(
             &from_int);
```
Should we go with that instead?

james-d-mitchell: Nah, the lambda is better. Sorry for the noise.

Joseph-Edwards:
This seems to be failing because we are getting values for `LIMIT_MAX` that are not successfully being compared:
```py
In [1]: from libsemigroups_pybind11 import *

In [2]: gens = [
   ...:     Perm([1, 0, 2, 3, 4] + list(range(5, 255))),
   ...:     Perm([1, 2, 3, 4, 0] + list(range(5, 255))),
   ...: ]

In [3]: S = SchreierSims(gens)

In [4]: x = list(S.sift(gens[0]))

In [5]: y = list(S.one())

In [6]: a, b = next((a, b) for a, b in zip(x,y) if a != b)

In [7]: a, b
Out[7]: (LIMIT_MAX, LIMIT_MAX)

In [8]: a == b
Out[8]: False
```

james-d-mitchell:
> This seems to be failing because we are getting values for `LIMIT_MAX` that are not successfully being compared:
> 
> ```py
> 
> In [1]: from libsemigroups_pybind11 import *
> 
> 
> 
> In [2]: gens = [
> 
>    ...:     Perm([1, 0, 2, 3, 4] + list(range(5, 255))),
> 
>    ...:     Perm([1, 2, 3, 4, 0] + list(range(5, 255))),
> 
>    ...: ]
> 
> 
> 
> In [3]: S = SchreierSims(gens)
> 
> 
> 
> In [4]: x = list(S.sift(gens[0]))
> 
> 
> 
> In [5]: y = list(S.one())
> 
> 
> 
> In [6]: a, b = next((a, b) for a, b in zip(x,y) if a != b)
> 
> 
> 
> In [7]: a, b
> 
> Out[7]: (LIMIT_MAX, LIMIT_MAX)
> 
> 
> 
> In [8]: a == b
> 
> Out[8]: False
> 
> ```

That seems all sorts of wrong, can we discuss tomorrow am?

james-d-mitchell (approved):
Looks good any reason to not merge?

PR #420, opened 2026-04-30
Doc fixes

james-d-mitchell:
A bunch of doc fixes many coming from codex.

PR #422, opened 2026-04-30
Doc fixes 2

james-d-mitchell:
This a bunch more doc fixes found (mostly) by Codex disjoint from #420


src/knuth-bendix.cpp
```diff
@@ -259,13 +259,13 @@ non-trivial classes of *kb1* in *kb2* if there are only finitely many finite
 such classes, regardless of whether or not *kb2* or *kb1* has infinitely many
 classes.
 
-:param kb1: the instance to compute the partition.
+:param kb1: the instance to compute the partition with.
```
Joseph-Edwards, line 262: I'm not sure if "with" is the right word here, but I don't have any other suggestions I'm convinced are better. "the instance used to compute the partition of \*kb2\*"?

Joseph-Edwards (commented):
I think the changes in this PR are good. I noticed there were a few things that had been corrected in some files but not others (e.g. all return sentences begin with a capital letter, but param sentences do not) so I've added a few more commits to catch a few more. I have one question below, but otherwise I'm happy to merge, provided you're happy with my new changes @james-d-mitchell?

src/knuth-bendix.cpp
james-d-mitchell: I see where you're coming from, but I think adding "with" here is a good change (or possibly neutral). I think of this in the following way: both `kb1` and `kb2` represent equivalence relations, we are trying to partition the normal forms related to `kb2` into their classes in `kb1`, which could be described as "partitioning the normal forms of `kb2` with `kb1`". Does this sound convincing?

Joseph-Edwards: Yeah that explanation makes sense. Thanks!

PR #431, opened 2026-05-10
Add commutator functions

### LINKED ISSUE #428: Add bindings for `Commutator` and `add_commutator_rules`
In light of https://github.com/libsemigroups/libsemigroups/pull/973, we should add the `commutator` and `add_commutator_rule`.

Joseph-Edwards:
This PR adds bindings and tests for the functions `commutator` and `add_commutator_rule`.

There was a slight hiccup here, because Pybind11 binds C++ chars as Python strings (with a few extra steps). This was causing some issues with overload resolution for the functions with the optional `id` parameter. Specifically, does `add_commutator_rule(p, x, y, z)` mean "add the commutator rule $[x, y] = \epsilon$ with inverses given by `z`" or does it mean "add the commutator rule $[x, y] = z$ with inverses inferred from the presentation"? It is reasonable to want to be able to do either of these things, but difficult to express your intent to Python, because `id` has the same type as the other parameters.

To get around this, I've made the parameter `id` a keyword-only argument (see [PEP 3102](https://peps.python.org/pep-3102/) for more info). Therefore, if we want to specify a non-standard identity, we need to explicitly provide it as a keyword argument (e.g. `add_commutator_rule(p, x, y, id=z)`). I've tried to make this clear in the doc, but if I'm happy to add more detail if it's needed.

I don't think we've added any keyword-only arguments before, so I needed to make a few tweaks to some of our scripts and helpers to accommodate this.

### DIFF
--- a/docs/source/data-structures/presentations/present-helpers.rst
+++ b/docs/source/data-structures/presentations/present-helpers.rst
@@ -34,6 +34,7 @@ Contents
 .. autosummary::
     :signatures: short
 
+    add_commutator_rule
     add_cyclic_conjugates
     add_identity_rules
     add_inverse_rules
@@ -43,6 +44,7 @@ Contents
     are_rules_sorted
     balance
     change_alphabet
+    commutator
     contains_rule
     first_unused_letter
     greedy_reduce_length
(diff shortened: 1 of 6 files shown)


james-d-mitchell:
Thanks @Joseph-Edwards, I'm happy with the keyword argument stuff you've proposed (in fact, in light of our recent conversation about this, I think maybe `alphabet` and `inverses` should be keyword args too. Obviously the CI failing needs to be fixed, but in principle I think this is good.

PR #434, opened 2026-05-14
Dot updates

james-d-mitchell:
Adds the functionality in 

https://github.com/libsemigroups/libsemigroups/pull/985

to the python bindings

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -50,6 +50,67 @@ file and render it with the Graphviz_ installation on your system.
 )pbdoc");
 
     ////////////////////////////////////////////////////////////////////////////
+    // Attr enum
+    ////////////////////////////////////////////////////////////////////////////
+
+    py::options options;
+    options.disable_enum_members_docstring();
+
+    py::enum_<Dot::Attr> attr(dot, "Attr", R"pbdoc(
+The values in this enum can be used to indicate how attribute values are
+written in the DOT_ output.
+
+The valid values are:
+
+.. py:attribute:: Attr.string
+  :value: <Attr.string: 0>
+
+  Value indicating that the attribute value should be written as a quoted
+  string.
+
+.. py:attribute:: Attr.html
+  :value: <Attr.html: 1>
+
+  Value indicating that the attribute value should be written as an HTML-like
+  label.
+)pbdoc");
+    attr.value("string", Dot::Attr::string).value("html", Dot::Attr::html);
+    attr.def("__repr__", [](Dot::Attr const& val) {
+      return to_human_readable_repr(val, ".");
+    });
+
+    ////////////////////////////////////////////////////////////////////////////
+    // Kind enum
+    ////////////////////////////////////////////////////////////////////////////
+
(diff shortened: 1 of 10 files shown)


src/dot.cpp
```diff
@@ -49,6 +49,67 @@ its DOT_ source code string (:any:`Dot.to_string`). Write the source code to a
...
+.. py:attribute:: Attr.string
+  :value: <Attr.string: 0>
+
+  Value indicating that the attribute value should be written as a quoted
+  string.
+
+.. py:attribute:: Attr.html
+  :value: <Attr.html: 1>
+
+  Value indicating that the attribute value should be written as an HTML-like
```
Joseph-Edwards, line 74:
```suggestion
  Value indicating that the attribute value should not be written as an HTML-like
```

Joseph-Edwards (changes requested):
This all seems good (modulo one small typo below). I will be happy to merge this once the upstream is merged and the CI passes.

src/dot.cpp
james-d-mitchell: Unless I'm losing my mind (possible), I think it was correct the first time, using `Attr.html` indicates that it should be written as an html-like value. Not sure?

Joseph-Edwards: Ah yes, you're exactly right; my bad! In `libsemigroups` there was a typo in the analogous docstring, and I must've read this one too quickly.

james-d-mitchell: Phew, this at least is not evidence of my mental decline

tests/test_froidure_pin.py
```diff
@@ -161,13 +162,50 @@ def check_idempotents(S):
 def check_cayley_graphs(S):
     ReportGuard(False)
 
+    gen_names = "abcdefghijklmnopqrstuvwxyz"[: S.number_of_generators()]
+
+    d = froidure_pin.dot_current_right_cayley_graph(S)
+    assert isinstance(d, Dot)
+    assert d.kind() == Dot.Kind.digraph
+    assert len(d.nodes()) == S.current_size()
```
Joseph-Edwards: This assertion does not hold; `d.nodes()` also contains the nodes that represent the legend, so is bigger than `S.current_size()`:
```pycon
>>> from libsemigroups_pybind11 import Transf, FroidurePin, froidure_pin
>>> gens = [Transf([1, 0, 2]), Transf([1, 2, 0]), Transf([0, 0, 1])]
>>> S = FroidurePin(gens)
>>> print(froidure_pin.dot_current_right_cayley_graph(S).nodes())
[<dot node "0">, <dot node "1">, <dot node "2">, <dot node "cluster_legend_head">, <dot node "cluster_legend_tail">]
```

Is this the intended behaviour @james-d-mitchell?

james-d-mitchell: Yes it is the intended behavior the test is rubbish will update!

Joseph-Edwards:
Seems there is a slight issue with the doc. I'll fix that now then get this merged. Thanks @james-d-mitchell!

Joseph-Edwards:
Turns out this doc fix isn't quite as simple as I had hoped. Will keep investigating

PR #441, opened 2026-05-21
Refactor import behaviour

### LINKED ISSUE #433: `from libsemigroups_pybind11 import *` imports **everything**
When we try and install everything, we should actually install **everything**

Joseph-Edwards:
This PR makes changes to the package's import structure, and makes use of the `__all__` variable to define what gets imported during wildcard imports. In most cases, the objects that have been explicitly named in `__all__` correspond to the objects that were already implicitly included when performing wildcard imports. However, the process of constructing the `__all__` lists revealed some small issues that this PR also resolves. These are listed below.

- The following classes that had previously been omitted are now included in `from libsemigroups_pybind11 import *`:
  - `PathsFromRoots`,
  - `PathsToRoots`,
  - `Perm16`,
  - `PPerm16`,
  - `PTransf16`,
  - `Transf16`, and
  - `Vect16`.
- Shadowing conflicts in the following objects have been resolved:
  - `Dot` from `libsemigroups_pybind11.detail` vs `_libsemigroups_pybind11`,
  - `is_obviously_infinite` submodule vs function, and
  - `to` function vs submodule.
- The `detail` submodule is not imported by `libsemigroups_pybind11 import *`.
- The `DISCLAIMER` variable is not imported by `from libsemigroups_pybind11 import *`.
- The `examples` submodule is now accessible from the `presentation` submodule without having to explicitly import it. That is, the following now works:
```pycon
>>> from libsemigroups_pybind11 import *
>>> presentation.examples.stylic_monoid(10)
<monoid presentation with 10 letters, 340 rules, and length 2,010>
```

Resolves #433.


james-d-mitchell:
Apart from all the test failures this is awesome and I look forward to merging it!

PR #444, opened 2026-05-27
Rename `parse_relations` to `parse`

### LINKED ISSUE #427: `parse_relations` is a terrible name for the function
Because it only parses a single word, not relations at all. We should rename it to `parse_word` or something. This is related to:

https://github.com/libsemigroups/libsemigroups/issues/968

(i.e. if that issue is resolved, then we can leave this function name the same!).

Joseph-Edwards:
This PR deprecates the function `parse_relations` and adds the function `parse`. Other than their names, these two functions are identical.

Closes #427.


james-d-mitchell:
Very good thanks @Joseph-Edwards

PR #446, opened 2026-05-29
Add `to_ace_string`

Joseph-Edwards:
This PR adds a function that can convert a presentation into ACE input. I am not 100% happy with the documentation here, because the function behaves differently (specifically, throws errors in different circumstances) dependent on whether it is called on a presentation over strings or words. I've tried to give a general idea of what might happen, but I'm keen to hear any suggestions on how to make it better. Should I be more verbose about what happens in the two cases, or is it fine as it is?

### DIFF
--- a/src/present.cpp
+++ b/src/present.cpp
@@ -1615,8 +1615,8 @@ are created by taking quotients of free semigroups or monoids.
 :only-document-once:
 Return the code that would create a presentation to be used with ACE.
 
-This function returns the string of ACE input that could be used to create an
-object with the same alphabet and rules as *p*.
+This function returns the string of ACE :cite:`Havas1999iy` input that could be
+used to create an object with the same alphabet and rules as *p*.
 
 :param p: the presentation.
 :type p: Presentation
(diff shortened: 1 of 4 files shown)


src/present.cpp
```diff
@@ -1605,6 +1605,38 @@ are created by taking quotients of free semigroups or monoids.
...
 :rtype: str
+)pbdoc");
+      m.def(
+          "presentation_to_ace_string",
+          [](Presentation_ const& p) { return presentation::to_ace_string(p); },
+          py::arg("p"),
+          R"pbdoc(
+:sig=(p: Presentation) -> str:
+:only-document-once:
+Return the code that would create a presentation to be used with ACE.
```
james-d-mitchell:
```suggestion
Returns a string containing the content of an ACE input file.
```

james-d-mitchell: Also link to ACE x lots in this doc

src/present.cpp
```diff
@@ -1605,6 +1605,38 @@ are created by taking quotients of free semigroups or monoids.
...
+:raises LibsemigroupsError: if any of the letters in the alphabet of
+  *p* are not lowercase, or do not correspond to a lowercase letter when
+  transformed to a string using :any:`words.human_readable_letter`.
+
+.. note::
+
+  ACE assumes that presentations are group presentations, where the the alphabet
+  consists of lowercase letters, and inverses correspond to the equivalent
+  uppercase letters. Therefore, a valid ACE presentation may not be a valid
+  libsemigroups presentation; for example, it is possible for relations in an
```
james-d-mitchell: libsemigroups should be a link and so should "presentation"

james-d-mitchell:
I think this looks good apart from the changes suggested.

src/present.cpp
Joseph-Edwards: I can add a reference to our bibliography entry for ACE, but  haven't been able to find a good link for ACE. We use [https://staff.itee.uq.edu.au/havas/](https://staff.itee.uq.edu.au/havas/) in several places, but when I try and follow this link I get met with "502: Bad Gateway".

james-d-mitchell: I also investigated this, but I think the ACE webpage is no longer available, so there's nothing to reference directly. Let's just leave it as is, maybe with a comment that we aren't able to locate anything to cite, and maybe a link to the GAP package which includes a copy of the source code of the ACE standalone IIRC.

PR #449, opened 2026-06-08
Update KnuthBendix

james-d-mitchell:


james-d-mitchell:
This will fail until

https://github.com/libsemigroups/libsemigroups/pull/1004

is merged.

src/libsemigroups_pybind11/to.py
```diff
@@ -40,6 +47,7 @@
 from .todd_coxeter import ToddCoxeter as _ToddCoxeter
 
 
+# FIXME this function converts list[int] -> list, but shouldn't
```
Joseph-Edwards, line 50: I will open an issue for this and fix in due course

Joseph-Edwards: Never mind; that issue is already open! I will still fix in due course.

Joseph-Edwards (approved):
All looks good to me. Thanks @james-d-mitchell!

PR #453, opened 2026-06-18
Knuth bendix search

james-d-mitchell:


james-d-mitchell:
TODO:

- [ ] `copy` and `__copy__`
- [ ] use `to_human_readable_repr` in `__repr__` 

both of these require changes to libsemigroups itself first.

james-d-mitchell:
https://github.com/libsemigroups/libsemigroups/pull/1019

src/knuth-bendix.cpp
```diff
@@ -448,13 +447,471 @@ redundant in this way, then ``None`` is returned.
...
+)pbdoc");
+
+      thing.def(
+          "depth_max",
+          [](TietzeExplorer_& self, size_t val) -> TietzeExplorer_& {
+            return self.depth_max(val);
+          },
+          py::arg("val"),
+          R"pbdoc(
+:sig=(self: TietzeExplorer, val: SupportsInt | SupportsIndex) -> TietzeExplorer:
```
Joseph-Edwards, line 552: I think we decided on using `int` instead of `SupportsInt | SupportsIndex`.

Joseph-Edwards (commented):
Other than a comment about type hints, I think this is good to go. Thanks @james-d-mitchell

james-d-mitchell:
We decided to go in a different direction in libsemigroups itself, so the changes in this PR are no longer required.

PR #456, opened 2026-06-18
Alphabet

james-d-mitchell:


src/alphabet.cpp
```diff
@@ -0,0 +1,633 @@
...
+      // the alphabet.
+      //
+      // Example
+      // -------
+      //
+      // .. doctest:: python
+      //
+      //   >>> from libsemigroups_pybind11 import Alphabet
+      //   >>> Alphabet("abc").throw_if_letter_not_in_alphabet("a") is None
+      //   True)pbdoc");
```
Joseph-Edwards, line 520: Is this commented out deliberately?

tests/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+
+from libsemigroups_pybind11 import ReportGuard
+
+
+@pytest.fixture(autouse=True)
+def setup_each_test():
+    """Avoid having to write ReportGuard(false) in every test case"""
+    rg = ReportGuard(False)
+    yield
+    del rg
```
Joseph-Edwards, line 13:
```suggestion
    return ReportGuard(False)
```
?

Joseph-Edwards (commented):
Other than some small comments, this looks good to me! Thanks @james-d-mitchell

src/alphabet.cpp
```diff
@@ -0,0 +1,633 @@
...
+
+      ////////////////////////////////////////////////////////////////////////
+      // Constructors + initializers
+      ////////////////////////////////////////////////////////////////////////
+
+      thing.def(py::init<>(), R"pbdoc(
+:sig=(self: Alphabet, *, word: type) -> None:
+::only-document-once:
+Default constructor.
+This function constructs an empty alphabet.
```
Joseph-Edwards: This is not a default constructor because it has a parameter.

tests/conftest.py
Joseph-Edwards: Whilst looking into this, I realised our `doctest` reporting suppression hasn't been working. I think the fix is
```diff
 # Python code that is treated like it were put in a testsetup directive for
 # every file that is tested, and for every group.
 doctest_global_setup = """from libsemigroups_pybind11 import ReportGuard
-ReportGuard(False)"""
+rg = ReportGuard(False)"""
``` 
in line 192 of `docs/source/conf.py`.

If you make the `pytest` change above, please could you make the `doctest` one too. Thank you

src/alphabet.cpp
james-d-mitchell: Yeah I don't think this is necessary but since the code was generated thought I'd leave it here in case it was needed later

tests/conftest.py
james-d-mitchell: > ```suggestion
>     return ReportGuard(False)
> ```
> ?

I don't think this works, due to the lifetime of the ReportGuard ending before the tests are run. I might be wrong, but I tried a number of different things here and this one works. The `del` is unnecessary but stops the linter from complaining about `rg` being unused so I left it in

james-d-mitchell: > Whilst looking into this, I realised our `doctest` reporting suppression hasn't been working. I think the fix is
> 
> ```diff
> 
>  # Python code that is treated like it were put in a testsetup directive for
> 
>  # every file that is tested, and for every group.
> 
>  doctest_global_setup = """from libsemigroups_pybind11 import ReportGuard
> 
> -ReportGuard(False)"""
> 
> +rg = ReportGuard(False)"""
> 
> ``` 
> 
> in line 192 of `docs/source/conf.py`.
> 
> 
> 
> If you make the `pytest` change above, please could you make the `doctest` one too. Thank you

I did this already 😀

src/alphabet.cpp
james-d-mitchell: Done

PR #463, opened 2026-08-05
Update output relating to RPO

Joseph-Edwards:
In https://github.com/libsemigroups/libsemigroups/pull/1035, we swapped the meaning of `rpo` and `rev_rpo` for so that these orders are special cases of `wr_order` and `rev_wr_order` respectively. In this PR, we change the test output to match the new order behaviour.


jswent (approved):
lgtm, thanks!

PR #469, opened 2026-08-14
Add support for `wr_cmp`

james-d-mitchell:
This was introduced into `libsemigroups` in:

https://github.com/libsemigroups/libsemigroups/pull/1031

I haven't added bindings for `WrCmp` because we don't have bindings for the other order "objects". It is probably worthwhile opening an issue about that since they'd be useful for things like `WrCmp` and `WtLenLex` which have data.


james-d-mitchell:
This initially contained some changes that undid recent merged changes in `main` so I've removed those again!

Joseph-Edwards (approved):

PR #367, opened 2025-12-01
Sync weighted comparison operators

jswent:

PR #472, opened 2026-08-14
order: add `wt_lex_cmp` + `wt_lenlex_cmp`

james-d-mitchell:
Supercedes #367.

### DIFF
--- a/src/order.cpp
+++ b/src/order.cpp
@@ -358,8 +358,8 @@ lexicographically.
 .. doctest:: python
 
   >>> from libsemigroups_pybind11 import wt_lenlex_cmp
-  >>> wt_lenlex_cmp([1, 2], [1], [0, 0])
-  True
+  >>> wt_lenlex_cmp([1, 3], [1], [0, 0])
+  False
 )pbdoc");
 
       m.def(
@@ -399,8 +399,8 @@ lexicographically according to *alphabet*.
 .. doctest:: python
 
   >>> from libsemigroups_pybind11 import Alphabet, wt_lenlex_cmp
-  >>> wt_lenlex_cmp(Alphabet("ba"), [1, 1], "b", "a")
-  True
+  >>> wt_lenlex_cmp(Alphabet("ba"), [3, 1], "b", "a")
+  False
 )pbdoc");
 
       m.def(
@@ -433,8 +433,8 @@ Words are first ordered by their total weight and then lexicographically.
 .. doctest:: python
 
   >>> from libsemigroups_pybind11 import wt_lex_cmp
-  >>> wt_lex_cmp([1, 2], [1], [0, 0])
-  False
+  >>> wt_lex_cmp([1, 1], [1], [0, 0])
+  True
 )pbdoc");
 
       m.def(
@@ -474,7 +474,7 @@ first ordered by their total weight and then lexicographically according to
 .. doctest:: python
 
   >>> from libsemigroups_pybind11 import Alphabet, wt_lex_cmp
-  >>> wt_lex_cmp(Alphabet("ba"), [1, 1], "b", "a")
+  >>> wt_lex_cmp(Alphabet("ba"), [3, 1], "a", "b")
   True
 )pbdoc");
 
--- a/src/order.cpp
+++ b/src/order.cpp
@@ -175,6 +175,7 @@ compared by their positions in *alphabet*.
 :type y: str | list[int]
 :returns: Whether *x* is less than *y*.
 :rtype: bool
+
 :raises LibsemigroupsError: if either word contains a letter that does not
   belong to *alphabet*.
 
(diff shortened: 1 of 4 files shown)


src/order.cpp
```diff
@@ -137,6 +138,157 @@ word contains a letter that does not belong to *alphabet*.
...
+:type y: str | list[int]
+:returns: Whether *x* is less than *y*.
+:rtype: bool
+
+:raises LibsemigroupsError: if a letter is not a valid index into *weights*.
+
+.. doctest:: python
+
+  >>> from libsemigroups_pybind11 import wt_lenlex_cmp
+  >>> wt_lenlex_cmp([1, 2], [1], [0, 0])
```
Joseph-Edwards:
```suggestion
  >>> wt_lenlex_cmp([1, 3], [0, 0], [1])
```
It might be more instructive if this example uses words that do not have the same order when using `lenlex_cmp`.

src/order.cpp
```diff
@@ -137,6 +138,157 @@ word contains a letter that does not belong to *alphabet*.
...
+:returns: Whether *x* is less than *y*.
+:rtype: bool
+
+:raises LibsemigroupsError: if a letter does not belong to *alphabet*, or its
+  position in *alphabet* is not a valid index into *weights*.
+
+.. doctest:: python
+
+  >>> from libsemigroups_pybind11 import Alphabet, wt_lenlex_cmp
+  >>> wt_lenlex_cmp(Alphabet("ba"), [1, 1], "b", "a")
```
Joseph-Edwards: Same comment as above.

src/order.cpp
```diff
@@ -137,6 +138,157 @@ word contains a letter that does not belong to *alphabet*.
...
+:type y: str | list[int]
+:returns: Whether *x* is less than *y*.
+:rtype: bool
+
+:raises LibsemigroupsError: if a letter is not a valid index into *weights*.
+
+.. doctest:: python
+
+  >>> from libsemigroups_pybind11 import wt_lex_cmp
+  >>> wt_lex_cmp([1, 2], [1], [0, 0])
```
Joseph-Edwards: Same comment as above.

src/order.cpp
```diff
@@ -137,6 +138,157 @@ word contains a letter that does not belong to *alphabet*.
...
+:rtype: bool
+
+:raises LibsemigroupsError: if a letter does not belong to *alphabet*, or its
+  position in *alphabet* is not a valid index into *weights*.
+
+.. doctest:: python
+
+  >>> from libsemigroups_pybind11 import Alphabet, wt_lex_cmp
+  >>> wt_lex_cmp(Alphabet("ba"), [1, 1], "b", "a")
+  True
```
Joseph-Edwards, line 478: Same comment as above

tests/test_order.py
```diff
@@ -73,3 +75,41 @@ def test_deprecated_comparisons(old_compare, new_compare):
...
+    with pytest.raises(LibsemigroupsError):
+        compare(weights, [0], [3])
+
+
+def test_weighted_comparisons_use_different_tie_breakers():
+    """Check the length and lexicographic tie breakers differ."""
+    weights = [1, 2]
+
+    assert wt_lenlex_cmp(weights, [1], [0, 0])
+    assert not wt_lex_cmp(weights, [1], [0, 0])
```
Joseph-Edwards: There are three potential tie breakers (weight, length, lex) but this test only checks one of them (length). Please could you either rename this test or add test cases that use all three tie breakers.

tests/test_order.py
```diff
@@ -73,3 +75,41 @@ def test_deprecated_comparisons(old_compare, new_compare):
...
+@pytest.mark.parametrize("compare", [wt_lenlex_cmp, wt_lex_cmp])
+def test_weighted_comparisons_for_integer_words(compare):
+    """Check weighted comparisons and validation for integer words."""
+    weights = [2, 1, 6]
+
+    assert compare(weights, [0, 1], [2])
+    assert not compare(weights, [2], [0, 1])
+
+    with pytest.raises(LibsemigroupsError):
+        compare(weights, [0], [3])
```
Joseph-Edwards, line 118: Please also add a string case

tests/test_order.py
```diff
@@ -73,3 +75,41 @@ def test_deprecated_comparisons(old_compare, new_compare):
...
+    assert compare(alphabet, [1, 1], "b", "a")
+
+    word_alphabet = Alphabet([1, 0])
+    assert compare(word_alphabet, [1, 1], [1], [0])
+
+    with pytest.raises(LibsemigroupsError):
+        compare(alphabet, [1, 1], "c", "a")
+
+    with pytest.raises(LibsemigroupsError):
+        compare(alphabet, [1], "a", "b")
```
Joseph-Edwards, line 150: If we are going for complete coverage, we should add tests that check `wt_lenlex_cmp` with an alphabet can correctly order two words when the tie-breaker is length, and that exceptions are thrown when the input words are invalid `list[int]`s. However, this is arguably done by the C++ tests, so maybe isn't necessary here.

Joseph-Edwards (commented):
Some minor comments, but otherwise looks good. Thanks!

src/order.cpp
james-d-mitchell: Because of the rebase I'm making these changes locally

tests/test_order.py
james-d-mitchell: Yeah we're really just testing the interface not the underlying functions at all.

PR #474, opened 2026-08-17
Add order objects

### LINKED ISSUE #470: Bind `LenLexCmp` and the other order objects
These are currently unbound, mostly because the data-less versions (`LenLexCmp`,`LexCmp`) aren't really useful in python. They start being useful in `WrCmp` and `WtLenLexCmp` etc when they have data.

james-d-mitchell:


docs/source/data-structures/order/index.rst
```diff
@@ -10,40 +10,31 @@
 Orders
 ======
 
-This page contains the documentation for several functions for comparing words
-or strings with respect to certain reduction orderings.
+This page contains the documentation for classes and several functions for
+comparing words or strings with respect to certain reduction orderings.
```
Joseph-Edwards: I don't think we make a distinction between words and strings elsewhere. This should either be "strings or lists of ints" or "words"

docs/source/data-structures/order/functions.rst
```diff
@@ -0,0 +1,74 @@
+..
+    Copyright (c) 2024-2026 Joseph Edwards, J. D. Mitchell
```
Joseph-Edwards:
```suggestion
    Copyright (c) 2026 Joseph Edwards, J. D. Mitchell
```

docs/source/data-structures/order/index.rst
```diff
@@ -10,70 +10,32 @@
...
-
-.. autofunction:: rpo_cmp
-
-.. autofunction:: wr_cmp
-
-.. autofunction:: wt_lenlex_cmp
-
-.. autofunction:: wt_lex_cmp
+The classes and functions in ``libsemigroups_pybind11`` for ordering words and
+strings are described on the following pages:
```
Joseph-Edwards: Same comment about "words and strings"

src/libsemigroups_pybind11/du_narendran_rusinowitch.py
```diff
@@ -6,6 +6,7 @@
 
 """Subpackage for importing and wrapping :any:`du_narendran_rusinowitch`."""
 
+# pylint: disable=no-name-in-module
```
Joseph-Edwards, line 9: I don't think this should be necessary

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+  True
+  >>> LenLexCmp(Alphabet("ba"))("b", "a")
+  True
+)pbdoc");
+
+      thing.def(py::init<Alphabet<Word> const&>(),
+                py::arg("alphabet"),
+                R"pbdoc(
+:sig=(self: LenLexCmp, alphabet: Alphabet) -> None:
+Construct a len-lex comparison object from an alphabet.
```
Joseph-Edwards:
```suggestion
Construct a lenlex comparison object from an alphabet.
```

This change should be made throughout this PR too.

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+The values in this enum can be used as the arguments for functions such as
+:any:`ToddCoxeter.standardize` or :any:`WordRange.order` to specify which
+ordering should be used.
+
+The values :any:`Order.shortlex` and :any:`Order.recursive` are retained for
+backwards compatibility; use :any:`Order.lenlex` and :any:`Order.rev_rpo`,
+respectively, in new code.
+
+.. py:attribute:: Order.none
+  :value: <Order.left: 0>
```
Joseph-Edwards:
```suggestion
  :value: <Order.none: 0>
```

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+
+   No ordering
+
+.. py:attribute:: Order.lenlex
+  :value: <Order.right: 1>
```
Joseph-Edwards:
```suggestion
  :value: <Order.lenlex: 1>
```

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+
+    The lenlex ordering. Words are first ordered by length, and then
+    lexicographically.
+
+.. py:attribute:: Order.shortlex
+  :value: <Order.left: 2>
```
Joseph-Edwards:
```suggestion
  :value: <Order.lenlex: 1>
```

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+
+    The short-lex ordering. Words are first ordered by length, and then
+    lexicographically.
+
+    This is deprecated; use :any:`Order.lenlex` instead.
+
+.. py:attribute:: Order.lex
+  :value: <Order.right: 3>
```
Joseph-Edwards:
```suggestion
  :value: <Order.lex: 2>
```

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+
+    The lexicographic ordering. Note that this is not a well-order, so there
+    may not be a lexicographically least word in a given congruence class of
+    words.
+
+.. py:attribute:: Order.rpo
+  :value: <Order.left: 4>
```
Joseph-Edwards:
```suggestion
  :value: <Order.rpo: 3>
```

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+
+    The recursive-path ordering, as described in :cite:`Jantzen2012aa`
+    (Definition 1.2.14, page 24).
+
+.. py:attribute:: Order.rev_rpo
+  :value: <Order.right: 5>
```
Joseph-Edwards:
```suggestion
  :value: <Order.rev_rpo: 4>
```

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+
+    The reversed recursive-path ordering, based on the description in
+    :cite:`Jantzen2012aa` (Definition 1.2.14, page 24), where words are read
+    right-to-left before ordering.
+
+.. py:attribute:: Order.recursive
+  :value: <Order.right: 6>
```
Joseph-Edwards: Interestingly, my local install gives this:
```
In [8]: Order.recursive
Out[8]: <Order.rev_rpo: 4>
```
This one will almost certainly change (and at the very least will be reviewed), so this one probably doesn't really matter.

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+    (Definition 1.2.14, page 24).
+
+    This is deprecated; use :any:`Order.rpo` instead.
+
+.. doctest:: python
+
+  >>> from libsemigroups_pybind11 import Order
+  >>> Order.lenlex
+  <Order.lenlex: 1>
+  >>> Order.shortlex == Order.lenlex
```
Joseph-Edwards: Add one for `Order.recursive too`?

Joseph-Edwards: Perhaps this would be better in a test files instead.

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+  True
+)pbdoc")
+        .value("none", Order::none)
+        .value("lenlex", Order::lenlex)
+        .value("shortlex", Order::lenlex)
+        .value("lex", Order::lex)
+        .value("rpo", Order::rpo)
+        .value("rev_rpo", Order::rev_rpo)
+        .value("recursive", Order::rev_rpo);
```
Joseph-Edwards, line 2915: This doesn't match what the doc above says, but this can be changed in the upcoming rev_rpo PR

tests/test_order.py
Joseph-Edwards, line 1: The lack of structure in this file makes it a bit difficult to check we're testing the API fully. It would be better if this file was split up into different sections per order, and within each section, there were tests for:
- with alphabet
- without alphabet
- copy and repr (object only, not `_cmp` functions)
- rejects bad constructor arguments (object only, not `_cmp` functions)

As far as I can tell, we aren't obviously missing anything, but I'm not confident in that.

src/order.cpp
```diff
@@ -1230,87 +1242,1856 @@ differences within one level are compared using lenlex ordering.
...
+*alphabet* and only accepts words with the same type as *alphabet*.
+
+.. note::
+  The constructor fixes whether this object is alphabet-aware (i.e. constructed
+  from an :any:`Alphabet` object). It also fixes the word type of an
+  alphabet-aware object. In particular, ``LexCmp()`` only accepts ``init()``
+  (which does nothing); it cannot be changed into an alphabet-aware object by
+  calling ``init(alphabet)``. Similarly,
+  ``LexCmp(alphabet).init(new_alphabet)`` requires *new_alphabet* to have the
+  same word type as *alphabet*.
```
Joseph-Edwards, line 1657: The sentence "`LexCmp()` only accepts `init()`" doesn't quite make sense to me. Is this saying "Objects constructed with `LexCmp()` can only be initialized by calling `init()`?

Joseph-Edwards: Separately (and more importantly), we don't bind the `init()` function with no argument.

Joseph-Edwards (changes requested):
A number of tweaks (mostly related to the doc and tests), but overall this looks really good! Thanks @james-d-mitchell

src/libsemigroups_pybind11/du_narendran_rusinowitch.py
james-d-mitchell: Let's discuss this, I'm also not sure why this is necessary but without it I get a warning when running pylint locally.

src/order.cpp
james-d-mitchell: Yeah I'd prefer to keep the deprecated things out of the manual examples.

src/order.cpp
james-d-mitchell: So, I've updated this so that it is self-consistent, even if it is not consistent (yet) with libsemigroups.

src/order.cpp
james-d-mitchell: I noticed that we do implement the `init()` for reduction orders taking levels or weights, but that there's no point in those existing (you'd have to call `init(levels)` or `init(weights)` again anyway) so I've removed those from the bindings.

src/order.cpp
Joseph-Edwards: In that case, should we move `>>> Order.shortlex == Order.lenlex` into the test files too?

src/order.cpp
```diff
@@ -1230,87 +1241,1822 @@ differences within one level are compared using lenlex ordering.
...
+   No ordering
+
+.. py:attribute:: Order.lenlex
+  :value: <Order.lenlex: 1>
+
+    The lenlex ordering. Words are first ordered by length, and then
+    lexicographically.
+
+.. py:attribute:: Order.shortlex
+  :value: <Order.lenlex: 2>
```
Joseph-Edwards:
```suggestion
  :value: <Order.lenlex: 1>
```

src/order.cpp
```diff
@@ -1230,87 +1241,1822 @@ differences within one level are compared using lenlex ordering.
...
+
+    The short-lex ordering. Words are first ordered by length, and then
+    lexicographically.
+
+    This is deprecated; use :any:`Order.lenlex` instead.
+
+.. py:attribute:: Order.lex
+  :value: <Order.lex: 3>
```
Joseph-Edwards:
```suggestion
  :value: <Order.lex: 2>
```

src/order.cpp
```diff
@@ -1230,87 +1241,1822 @@ differences within one level are compared using lenlex ordering.
...
+
+    The lexicographic ordering. Note that this is not a well-order, so there
+    may not be a lexicographically least word in a given congruence class of
+    words.
+
+.. py:attribute:: Order.rpo
+  :value: <Order.rpo: 4>
```
Joseph-Edwards:
```suggestion
  :value: <Order.rpo: 3>
```

src/order.cpp
```diff
@@ -1230,87 +1241,1822 @@ differences within one level are compared using lenlex ordering.
...
+
+    The recursive-path ordering, as described in :cite:`Jantzen2012aa`
+    (Definition 1.2.14, page 24).
+
+.. py:attribute:: Order.rev_rpo
+  :value: <Order.rev_rpo: 5>
```
Joseph-Edwards:
```suggestion
  :value: <Order.rev_rpo: 4>
```

src/order.cpp
```diff
@@ -1230,87 +1241,1822 @@ differences within one level are compared using lenlex ordering.
...
+
+    The reversed recursive-path ordering, based on the description in
+    :cite:`Jantzen2012aa` (Definition 1.2.14, page 24), where words are read
+    right-to-left before ordering.
+
+.. py:attribute:: Order.recursive
+  :value: <Order.rev_rpo: 5>
```
Joseph-Edwards:
```suggestion
  :value: <Order.rev_rpo: 4>
```

james-d-mitchell:
@Joseph-Edwards thanks for the review, I think that should be everything now.

PR #485, opened 2026-08-28
Update rpo and rev_rpo

### LINKED ISSUE #477: In `main` the documentation for `rpo_cmp` and `rev_rpo_cmp` currently disagrees with the documentation in libsemigroups
In libsemigroups, `rpo_cmp` to read words from right to left, while in the doc here it claims to read them from left to right. This should be fixed in one place or the other!

### LINKED ISSUE #478: The doc for `rpo_cmp` + `rev_rpo_cmp` should include the definitions

### LINKED ISSUE #480: Remove the perfomance warning for `rpo_cmp` and `rev_rpo_cmp`
After some benchmarking, we realised that `rpo_cmp` does not have notably worse performance than the other comparison functions. As such, we should remove the disclaimers that say the performance is worse.

Joseph-Edwards:
This PR updates the Python bindings to reflect the changes in https://github.com/libsemigroups/libsemigroups/pull/1056. It also adds several `...RPO...` objects, where previously only `...RevRPO...` objects existed.


src/du-narendran-rusinowitch.cpp
```diff
@@ -46,24 +46,26 @@ namespace libsemigroups {
...
+alphabet is empty if this fails, or if the alphabet was empty to begin
+with.
 
 :param p: the presentation whose rules are to be oriented.
 :type p: Presentation
 
-:returns: An alphabet order orienting every rule, or an empty word if none
-  exists.
+:returns: The ordered alphabet, if such an alphabet exists. Otherwise, the
+  empty alphabet.
```
james-d-mitchell, line 62: This isn't really a comment about this PR, but maybe this should return `None` if it can't do it.

james-d-mitchell:
@codex review

james-d-mitchell (approved):
Looks good to me, let's see what Codex "thinks"

src/libsemigroups_pybind11/knuth_bendix.py
james-d-mitchell, line 137: I think Codex means there are some of the new `_KnuthBendix.*` that should be registered here too.

src/du-narendran-rusinowitch.cpp
Joseph-Edwards: Do you think we should change the C++ code to return an optional, or should this be a Python only change?

PR #487, opened 2026-09-03
Add missing include to `paths.cpp`

Joseph-Edwards:
This PR adds a missing include, that was revealed by https://github.com/libsemigroups/libsemigroups/pull/1075.


Joseph-Edwards:
The CI will fail until the libsemigroups PR is merged.

PR #489, opened 2026-09-08
Validate Presentation constructor keyword arguments

james-d-mitchell:
`Presentation(word=str, typo=True)` silently ignores the unexpected keyword because its constructor never forwards keyword arguments to the base validator. The same behavior affects `InversePresentation` after #488.

Forward `**kwargs` to the base constructor and declare the allowed keyword as `("word",)`. Using a tuple makes validation check complete keyword names rather than accepting substrings such as `"w"` or `"or"`.

Add parameterized tests covering both presentation classes, both supported word types, unexpected keywords alongside `word`, unknown keyword names on their own, and valid keyword construction.

Validation: all 120 tests in `tests/test_present.py` and `tests/test_presentation_examples.py` pass using the source Python package with the existing Python 3.13 native build. Before the fix, 18 new regression cases failed and the four valid-construction cases passed. Ruff lint and format checks pass for both changed files.

This PR is based on #488 so its diff contains only the keyword validation fix and tests. It should be retargeted to `main` after #488 merges.

AI disclosure: OpenAI Codex authored the code changes, regression tests, and this PR description at the maintainer's request.

### DIFF
--- a/src/libsemigroups_pybind11/presentation/__init__.py
+++ b/src/libsemigroups_pybind11/presentation/__init__.py
@@ -125,7 +125,7 @@ class Presentation(_CxxWrapper):
         """
         # super().__init__ checks if there are unexpected kwargs,
         # and sets _cxx_obj if the unique argument is a cxx_obj of type in _all_wrapped_cxx_types
-        super().__init__(*args, optional_kwargs="word")
+        super().__init__(*args, optional_kwargs=("word",), **kwargs)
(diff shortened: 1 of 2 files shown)


james-d-mitchell (commented):
Looks good to me, what say ye @Joseph-Edwards ?

Joseph-Edwards (approved):

PR #490, opened 2026-09-08
Avoid temporary objects in copy bindings

### LINKED ISSUE #468: Avoid unnecessary copies in `__copy__` and `copy`
Similar to the changes made in #467

james-d-mitchell:
The remaining `copy()` and `__copy__()` bindings return C++ objects by value, leaving pybind11 to move or copy those temporaries into Python-owned storage. Construct copies directly with `std::make_unique<T>(self)`, following the SchreierSims fix in #467.

Update 98 bindings across 25 C++ files and add the required `<memory>` includes. Add tests for both WordGraph copy entry points that check independent mutation and continued use after the original is deleted. This preserves the Python API and does not add an allocation compared with pybind11's existing by-value conversion.

Fixes #468.

Validation in the `libsemigroups_pybind11_dev` mamba environment:
- Clean C++ extension rebuild, with HPCombi enabled.
- `make check` on this PR branch: 1,509 doctests and 678 pytest tests passed, using the rebuilt extension from the checkout.
- C++ lint, Ruff lint and formatting, and pylint checks passed for the changed files as applicable.
- A small compiled pybind11 check confirmed one copy plus one move for the by-value pattern, versus one copy and no move for direct heap construction. This is not a benchmark of the individual bound types.

AI disclosure: OpenAI Codex authored the code changes, tests, commit, and this PR description at the maintainer's request.

### DIFF
--- a/src/action.cpp
+++ b/src/action.cpp
@@ -23,6 +23,7 @@
 // TODO Left/RightActionPerm
 
 // C++ stl headers....
+#include <memory>       // for make_unique
 #include <string_view>  // for string_view
 
 // libsemigroups headers
@@ -115,10 +116,12 @@ In this documentation we refer to:
       // Action code.
       thing.def(py::init<>());
 
-      thing.def("__copy__", [](Action_ const& self) { return Action_(self); });
+      thing.def("__copy__", [](Action_ const& self) {
+        return std::make_unique<Action_>(self);
+      });
       thing.def(
           "copy",
-          [](Action_ const& self) { return Action_(self); },
+          [](Action_ const& self) { return std::make_unique<Action_>(self); },
           R"pbdoc(
 :sig=(self: Action) -> Action:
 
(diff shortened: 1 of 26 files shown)


Joseph-Edwards (commented):
It's not 100% clear to me what the issue was here, but that is a me problem. If including `make_unique` is the correct solution to the problem, then I am happy with the changes in this PR.

james-d-mitchell:
Thanks @Joseph-Edwards. Small explanation: when we do `return x;` inside the constructor, or the lambda calling the constructor, `x` is allocated on the stack (more or less, it depends on what `x` is, some or all of it may be allocated on the heap too). This value is then returned from the function, and moved into a `unique_ptr` which is seemingly how pybind11 stores things by default. Moving into the `unique_ptr` allocates on the heap. The issue that arose in `SchreierSims` was that this allocated a lot of memory on the stack, and this sometimes caused a seg fault in the CI. If we return a `unique_ptr`, then the object is initially allocated on the heap, and so there's no issue with stack overflows. It seems pretty unlikely to me that the other constructors changed in this PR would actually allocate lots of memory on the stack, `SchreierSims` is kind of special, but there's no harm in doing this, so why not?

PR #505, opened 2026-09-12
Codex/presentation rules view

james-d-mitchell:
A second attempt at resolving #391


src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
   namespace py = pybind11;
 
   namespace {
+    // Only Presentation.rules creates these non-owning views. The property
+    // getter keeps the presentation alive; independent results are Python
+    // lists. AI assistance: OpenAI Codex helped implement and test these
```
Joseph-Edwards: What does "independent results are Python lists" mean?

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+
+    struct RulesSlice {
+      py::ssize_t start, stop, step, length;
+
+      RulesSlice(py::slice const& slice, size_t size) {
+        if (!slice.compute(size, &start, &stop, &step, &length)) {
+          throw py::error_already_set();
+        }
+      }
+    };
```
Joseph-Edwards: Please could you add some comments that explain what `start`, `stop`, `step` and `length` are.

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+      using View   = RulesView<Word>;
+      using Vector = std::vector<Word>;
+      using Index  = py::ssize_t;
+
+      auto wrap_index = [](Index i, size_t size) {
+        if (i < 0) {
+          i += static_cast<Index>(size);
+        }
+        if (i < 0 || static_cast<size_t>(i) >= size) {
+          throw py::index_error();
```
Joseph-Edwards:
```suggestion
          throw py::index_error("list index out of range");
```
This seems to be the standard error message

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+               [wrap_index](View& self, Index i, Word const& word) {
+                 self.vector()[wrap_index(i, self.vector().size())] = word;
+               })
+          .def("__setitem__",
+               [](View& self, py::slice const& slice, Vector const& words) {
+                 RulesSlice indices(slice, self.vector().size());
+                 if (static_cast<size_t>(indices.length) != words.size()) {
+                   throw std::runtime_error("Left and right hand size of slice "
+                                            "assignment have different sizes!");
+                 }
```
Joseph-Edwards, line 233: Python list assignment with slices supports the left-hand and right-hand sides being different length:
```pycon
>>> x = ['a', 'b', 'c']

>>> x[0:2]
['a', 'b']

>>> x[0:2] = ['d', 'e', 'f']

>>> x
['d', 'e', 'f', 'c']
```
The doc below makes this requirement clear, so it's probably not that important if we don't support this here.

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+                   rules.erase(rules.begin() + indices.start,
+                               rules.begin() + indices.start + indices.length);
+                 } else {
+                   // Delete in descending index order so remaining indices stay
+                   // valid.
+                   for (Index i = indices.length; i > 0; --i) {
+                     rules.erase(rules.begin() + indices.start
+                                 + (i - 1) * indices.step);
+                   }
+                 }
```
Joseph-Edwards, line 271: We can remove the first two cases because they are handled by the code in the `else` block.

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+              "insert",
+              [](View& self, Index i, Word const& word) {
+                auto& rules = self.vector();
+                if (i < 0) {
+                  i += static_cast<Index>(rules.size());
+                }
+                if (i < 0 || static_cast<size_t>(i) > rules.size()) {
+                  throw py::index_error();
+                }
+                rules.insert(rules.begin() + i, word);
```
Joseph-Edwards, line 298: Much to my surprise, `insert` in Python does not throw an error if the index is out of bounds; it simply either appends to the front or the back if the index is too small or large, respectively.
```pycon
>>> x = ['a']

>>> x.insert(-100, 'b')

>>> x.insert(100, 'c')

>>> x
['b', 'a', 'c']
```

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+                return std::count(rules.begin(), rules.end(), word);
+              },
+              py::arg("x"))
+          .def(
+              "remove",
+              [](View& self, Word const& word) {
+                auto& rules = self.vector();
+                auto  it    = std::find(rules.begin(), rules.end(), word);
+                if (it == rules.end()) {
+                  throw py::value_error();
```
Joseph-Edwards: There should be an error message here. The normal Python one seems to be `TypeError: list.remove(x): x not in list`. I'm not sure if that is the most appropriate for our case, but words to that effect would be good.

src/present.cpp
```diff
@@ -47,6 +52,240 @@ namespace libsemigroups {
...
+          return false;
+        }
+        auto const sequence = other.cast<py::sequence>();
+        if (sequence.size() != self.vector().size()) {
+          return false;
+        }
+        for (size_t i = 0; i < sequence.size(); ++i) {
+          if (self.vector()[i] != sequence[i].cast<Word>()) {
+            return false;
+          }
```
Joseph-Edwards:
```pycon
>>> p = Presentation("ab")

>>> p.rules = ["aa", "bb"]

>>> p.rules == [[0, 0], [1, 1]]
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
----> 1 p.rules == [[0, 0], [1, 1]]

RuntimeError: Unable to cast Python instance of type <class 'list'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```

Provided it isn't too awkward, it would be good if were able to detect a mismatch in types and give a better error message. If we can't check the types directly, we could try wrapping this code in a try-catch block, catch `py::error_already_set` (I think, but not certain), and show our own error message.

src/present.cpp
```diff
@@ -72,18 +311,58 @@ available in the module :any:`libsemigroups_pybind11.presentation`.)pbdoc");
...
+              py::keep_alive<0, 1>()),
+          [](Presentation_& self, std::vector<Word> rules) {
+            self.rules = std::move(rules);
+          },
+          R"pbdoc(
+Mutable view of the rules of the presentation.
+
+The view supports indexing, slicing, iteration, comparison, and the methods
+``append``, ``extend``, ``insert``, ``pop``, ``clear``, ``count``, and ``remove``.
+Changes to the view change the presentation. Assigning an iterable to this
```
Joseph-Edwards:
```suggestion
Changes to the view change the the rules stored in the presentation. Assigning an iterable to this
```

Joseph-Edwards (changes requested):

james-d-mitchell:
Thanks @Joseph-Edwards those all make sense, will fix asap

src/present.cpp
Joseph-Edwards: Maybe our message should say `rules` rather than `list`

src/present.cpp
james-d-mitchell: Discussing this now, it seems likely that the final clause, should be replaced with a use of `std::remove_if` for better complexity.

src/present.cpp
james-d-mitchell: It does seem to be a `ValueError` for me, are you sure? @Joseph-Edwards

james-d-mitchell:
@Joseph-Edwards that's all the changes made now, happy to merge?

Issue #476, opened 2026-08-18
Add stub files for typing support


Joseph-Edwards:
A lot of the source code for this package is written in C++, and therefore it is not possible to provide inline typing support. To get around this, we could provide [stub files](https://typing.python.org/en/latest/spec/distributing.html) that include the typing information. This would have the following benefits:
- Users of our package could use static type checkers like mypy with `libsemigroups_pybind11`.
- We could use static type checkers whilst developing, which will likely identify some existing bugs.
- Better tab completion (I think).
- Sphinx can leverage information in stub files, so we may be able to get rid of some of the ~hacks~ bespoke software solutions relating to our docs.

PR #507, opened 2026-09-13
Add draft stub generation and Action typing

james-d-mitchell:
Dynamic wrapper methods and native bindings lose useful type information during ordinary stub generation. This adds a reusable generator that combines pybind11-stubgen discovery with source annotations, generic bases, and documented signatures.

- Require an explicit module and print a single stub to stdout, keeping diagnostics on stderr. Files are written only with `--output-dir`; replacing existing files requires `--force`.
- Default extension generation to publicly exported native classes and their typing dependencies. `--no-public-only` requests the full extension, and `--strict` stops on backend diagnostics.
- Add sample Action and BMat8 helper-module stubs, make Action generic in its element and point types, and correct the `copy_cxx_mem_fns` class annotation.
- Document usage and cover generation, public-class selection, output handling, and mypy consumer checks.

Generated stubs remain drafts: some native signatures still need manual correction. The mypy tests cover Action and an isolated BMat8 section, not the complete native stub. This does not add a `py.typed` marker.

Validation:

- `uv run --offline --no-sync --with pybind11-stubgen==2.5.5 --with mypy==2.3.1 pytest tests/test_generate_stubs.py tests/test_action.py tests/test_bmat8.py -k 'not test_action_transf'`: 58 passed, 1 deselected. The excluded test enumerates a large transformation action.
- Both pre-commit hook stages passed on all changed files; explicit Ruff checks and formatting passed for the generator as well.
- Confirmed tests use the package and built extension from this checkout. No C++ rebuild was needed; documentation builds and the full test suite were not run for this Python typing/tooling change.

AI disclosure: OpenAI Codex helped prepare the implementation, tests, documentation, commit message, and this PR description, and ran validation. The commit includes `Assisted-by: Codex <codex@openai.com>` and `Co-authored-by: Codex <codex@openai.com>` trailers.


james-d-mitchell:
This is a partial attempt (or start of an attempt) to resolve #476

